PR #1, opened 2020-03-06
[usermanager] Initial commit. Fixes JB#48597

mkenttala:


README.md
```diff
@@ -1 +1,67 @@
...
+Sailfish OS device users. The main system user is created during the first boot
+of the device. This user can the use this API to add additional users which do
+only have access to some of the available functionality and services.
+
+## API:
+
+Below is the D-Bus interface introspection:
+
+`
+    <!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
```
monich: I suggest to have a link to the actual file, otherwise this comment will have to be updated every time the spec changes and sooner or later someone will forget to do that.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+    if (!addGroup(user)) {
+        QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, "Creating user group failed"));
+        return;
+    }
+
+    struct lu_error *error = nullptr;
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, lu_prompt_console, NULL, &error);
+    if (!context) {
+        qWarning() << "Error creating context:" << lu_strerror(error);
+        QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, lu_strerror(error)));
```
inzanity: I think inheriting `QDBusContext` and using `sendErrorReply()` would look nicer, no need for the `message` argument (if the message is needed for other reasons, like authorization, `message()` can be used). Also the introspection data would be compatible with other systems as well.

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+<node>
+  <interface name="org.sailfishos.usermanager">
+    <method name="users">
+        <arg direction="in" type="(i)" name="message"/>
```
monich: Hmm... What kind of message is that and why is it array of 32-bit ints?

sailfishusermanager.cpp
monich: Yeah, this message thing looks like some obscure QDBus specific hackery. I would vote for taking it out, too.

monich: I would also vote for not using Qt in daemons unless it's really necessary, but that's obviously too much to ask :D

org.sailfishos.usermanager.xml
inzanity: It is actually a struct containing one int, but looks like some kind of qdbusxml2cpp wizardry.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+#include <pwd.h>
+#include <grp.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <libuser/user.h>
+
+const char* DBUS_INTERFACE = "org.sailfishos.usermanager";
+const char* DBUS_OBJECT_PATH = "/";
```
Tomin1: While we have quite a few components that have path /, it should be in the format /org/sailfishos/usermanager although it might sound a bit redundant. One reasoning here http://0pointer.de/blog/projects/versioning-dbus.html.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+    struct group *grent = getgrnam(USER_GROUP);
+    if (grent) {
+        for (int i = 0; grent->gr_mem[i]; i++) {
+            SailfishUser user;
+            user.user = grent->gr_mem[i];
+
+            struct passwd *pw = getpwnam(grent->gr_mem[i]);
+            if (pw && pw->pw_gecos)
+                user.name = pw->pw_gecos;
```
Tomin1, line 69: This should probably take the value of pw_gecos until first ",". That would be more compatible with chfn and typical use of gecos field. That also means we should validate that name doesn't contain , characters (I think removing them is okay).

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+    if (chown(destination.toLocal8Bit(), pw->pw_uid, pw->pw_gid)) {
+        qWarning() << "Home directory ownership change failed";
+        return false;
+    }
+    if (chmod(destination.toLocal8Bit(), HOME_MODE)) {
+        qWarning() << "Home directory permissions change failed";
+        return false;
+    }
+
+    foreach (const QString& file, source.entryList(QDir::Files | QDir::NoDotAndDotDot | QDir::Hidden)) {
```
Tomin1: I guess we don't have any use case for copying directories too?

org.sailfishos.usermanager.xml
Tomin1: Annotation says it's QDBusMessage but I don't think that's how they should be used.

rpm/user-managerd.spec
```diff
@@ -0,0 +1,39 @@
+Name: user-managerd
+Version: 0.0.1
+Release: 1
+Summary: Sailfish user manager daemon
+Group: System/Base
+License: BSD-3-Clause
```
Tomin1: Should be just BSD.
https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#SoftwareLicenses

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
+/*
+ * Copyright (c) 2020 Jolla Ltd.
+ * Copyright (c) 2020 Open Mobile Platform LLC.
+ *
+ * All rights reserverd.
```
Tomin1: Typo, extra r. Although, I think this should include the full BSD license text really.

sailfishusermanager.cpp
Tomin1: That would look nicer. This should also define its own errors instead of always returning QDBusError::Failed (org.freedesktop.DBus.Error.Failed) which tells nothing about what went wrong. There are also some possibly more appropriate errors to use in QDBusError like InvalidArgs or maybe InternalError if administrator can't do anything about it.

sailfishusermanager.h
```diff
@@ -0,0 +1,76 @@
...
+ * BSD 3-Clause License, see LICENSE.
+ */
+
+#ifndef SAILFISHLOGIND_H
+#define SAILFISHLOGIND_H
+
+#include <QObject>
+#include <QDBusArgument>
+#include <QDBusMessage>
+#include <libuser/error.h>
```
Tomin1: Some leftover?

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+#include <sys/stat.h>
+#include <libuser/user.h>
+
+const char* DBUS_INTERFACE = "org.sailfishos.usermanager";
+const char* DBUS_OBJECT_PATH = "/";
+const char* USER_GROUP = "users";
+const char* GROUPS_IDS_FILE = "/usr/lib/setup/group.ids";
+const char* GROUPS_SYSTEM = "SYSTEM_GROUPS";
+const char* GROUPS_USER = "USER_GROUPS";
+const char* SKEL_DIR = "/etc/skel";
```
Tomin1: Most (or all?) of these should use QStringLiteral I think.

Tomin1:
While not necessary and probably also somewhat lot of work, I think the code could benefit from refactoring libuser parts to their own class with some clean interfaces. It could hide some of the nastiness and leave the interface definitions leaner.

org.sailfishos.usermanager.xml
monich: > It is actually a struct containing one int, but looks like some kind of qdbusxml2cpp wizardry.

Your right, struct. That doesn't make it less weird, though.

sailfishusermanager.cpp
monich: The referenced page doesn't actually give any reasoning, it just says that it's a bad idea because "it makes it very hard to implement more than one service or service version on a single D-Bus connection" but doesn't explain why. I never noticed any difficulties with `/` as a path. AFAICT it's as easy and simple as using any other path.

sailfishusermanager.cpp
monich: I'm not sure, but it may even make sense to throw protocol specific errors, to let the client distinguish those from plain D-Bus errors. That may be an overkill, though - why would a client be interested in what exactly went wrong on the service side?

sailfishusermanager.cpp
spiiroin: Technically, yes, path is a path and does not matter. 

However, given that there is documented convention, following the convention is less "surprising" than deviating from it -> I'd reverse the question: What if any benefit there is from using "/"?

main.cpp
```diff
@@ -0,0 +1,19 @@
...
+ * BSD 3-Clause License, see LICENSE.
+ */
+
+#include <QCoreApplication>
+
+#include "sailfishusermanager.h"
+
+int main(int argc, char *argv[])
+{
+    QCoreApplication a(argc, argv);
```
spiiroin: qt style nit: "Avoid short or meaningless names (e.g. "a", "rbarr", "nughdeget")"

rpm/user-managerd.spec
```diff
@@ -0,0 +1,39 @@
+Name: user-managerd
+Version: 0.0.1
+Release: 1
+Summary: Sailfish user manager daemon
+Group: System/Base
```
spiiroin: We should not have "Group:" ?

rpm/user-managerd.spec
```diff
@@ -0,0 +1,39 @@
...
+%prep
+%setup -q -n %{name}-%{version}
+
+%build
+mkdir -p build && cd build
+%qmake5 DEFINES+='TARGET_ARCH=\\\"\"%{_target_cpu}\"\\\"' -recursive ..
+make %{?_smp_mflags}
+
+%install
+cd build && make INSTALL_ROOT=%{buildroot} install
```
spiiroin: nit: could be "make -C build ..."

sailfishusermanager.cpp
mkenttala: Well, my reasoning behind '/' is that since the path has no meaning for this interface that's the shortest one that works. I have no preference in this so whatever works for me.

org.sailfishos.usermanager.xml
mkenttala: It's not documented anywhere but qdbusxml2cpp requires those brackets for some reason, the type inside does not matter as it gets replaced.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+#include <QDir>
+
+#include <pwd.h>
+#include <grp.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <libuser/user.h>
+
+const char* DBUS_INTERFACE = "org.sailfishos.usermanager";
```
spiiroin: qt style nit: "For pointers or references, always use a single space between the type and '*' or '&', but no space between the '*' or '&' and the variable name"

sailfishusermanager.cpp
mkenttala: I added those QDBusMessage arguments to slots so that I could answer with an error, but I'll try that QDBusContext if I can get that to work. Otherwise need to ditch automatic qdbusxml2cpp.

sailfishusermanager.cpp
inzanity: For a more authoritative source for prefixing (or namepacing) object paths, it is recommended by FDo's [tutorial](https://dbus.freedesktop.org/doc/dbus-tutorial.html#objects).

sailfishusermanager.cpp
spiiroin: Or, even if they would remain c-strings (e.g. to facilitate qFatal("%s") use), using "const char foo[] ..." or "const char * const foo ..." might be more appropriate.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    if (!connection.registerService(DBUS_INTERFACE)) {
+        qFatal("Cannot register D-Bus service at %s", DBUS_INTERFACE);
+    }
+
+    new UsermanagerAdaptor(this);
+
+    m_timer = new QTimer(this);
+    connect(m_timer, SIGNAL(timeout()), qApp, SLOT(quit()));
```
spiiroin: nit: old style connect. could use connect(m_timer, &QTimer::timeout, qApp, &QCoreApplication::quit)?

rpm/user-managerd.spec
mkenttala: There's no 3-clause in that list but google found: https://spdx.org/licenses/BSD-3-Clause.html which does use this longer format for "short identifier". Anyway I copied this from some already existing package so I guess there's not yet convention for this :)

Tomin1: We should use Fedora's convention which is to use the short names on that list. See https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_field

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    return rv;
+}
+
+bool SailfishUserManager::addGroup(const QString &group)
+{
+    struct lu_error *error = nullptr;
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, lu_prompt_console, NULL, &error);
+    bool rv = true;
```
spiiroin: nit: this rv (returnValue?) is not used in case early return with literal false happens. in e.g. addUserToGroup() rv init is handled after such early returns - which I find slightly less confusing.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+        qWarning() << "Error creating context";
+        return false;
+    }
+
+    bool rv = true;
+    struct lu_ent *ent = lu_ent_new();
+    if (lu_group_lookup_name(context, group.toLocal8Bit(), ent, &error)) {
+        GValue *value = g_slice_new0(GValue);
+        g_value_init(value, G_TYPE_STRING);
+        g_value_set_string(value, user.toLocal8Bit());
```
spiiroin: These toLocal8Bit() things depend on having locale correctly set up. Would it be better / worse to use toUtf8() instead?

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+    if (!context) {
+        qWarning() << "Error creating context:" << lu_strerror(error);
+        return false;
+    }
+
+    struct lu_ent *ent_group = lu_ent_new();
+    if (lu_group_default(context, group.toLocal8Bit(), false, ent_group)) {
+        if (!lu_group_add(context, ent_group, &error)) {
+            qWarning() << "Adding group failed:" << lu_strerror(error);
```
spiiroin: There is lu_error_free() function, but no calls to it. Am I missing something or are these memory leaks?

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+        droidDefault.close();
+    }
+    QFile droidInet("/usr/lib/droid/hw-group.d/inet.list");
+    if (droidInet.open(QIODevice::ReadOnly)) {
+        while (!droidInet.atEnd()) {
+            line = droidInet.readLine();
+            if (!line.isEmpty()) {
+                if (!addUserToGroup(user, line.trimmed())) {
+                    droidInet.close();
+                    return false;
```
spiiroin: This return false on first failure logic -> a lot of things might have been despite the error. Should we care enough to attempt rollback? If not, would it be better/worse to do as much as possible before returning "it failed"?

sailfishusermanager.cpp
spiiroin: I have feeling that there is a ton of subtle gotchas related to this. Would there be some way to use existing code/scripts for skeleton setup?

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+        QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, "Could not find user"));
+    }
+
+    lu_ent_free(ent);
+    lu_end(context);
+}
+
+bool SailfishUserManager::removeDir(const QString &dir)
+{
+    QDir home(dir);
```
spiiroin: Using variable "home" in method that seems to be fully generic -> confusing.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+                if (!lu_group_modify(context, ent, &error)) {
+                    qWarning() << "Group modify failed:" << lu_strerror(error);
+                    QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, lu_strerror(error)));
+                    rv = false;
+                    break;
+                }
+            }
+        }
+    } else {
+        error = nullptr;
```
spiiroin: presumably done to avoid some lu_xxx() aborting because of already set error? but is this memory leak? should it be lu_error_free(&error) instead?

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+                qWarning() << "Could not find group";
+                // Error is not populated in this case so returning generic error
+                QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, "Could not find group"));
+            }
+        } else {
+            qWarning() << "User modify failed:" << lu_strerror(error);
+            QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, lu_strerror(error)));
+        }
+    } else {
+        qWarning() << "Could not find user";
```
spiiroin: matter of taste etc - feel free to ignore, but I find "dealing errors upfront" leads to more readable code than "stacking failure handling in else of else of else ..."

sailfishusermanager.cpp
```diff
@@ -0,0 +1,429 @@
...
+    lu_end(context);
+}
+
+void SailfishUserManager::removeUser(const QString& user, const QDBusMessage& message)
+{
+    m_timer->start();
+
+    if (!removeHome(user)) {
+        QDBusConnection::systemBus().send(message.createErrorReply(QDBusError::Failed, "Removing user home failed"));
+        return;
```
spiiroin, line 238: If both add and remove user functions bailout on first sign of troubles and leave semi valid user data behind, can we / how likely it is that we could end up in situation where some user names are permanently unusable?

sailfishusermanager.cpp
monich: That [tutorial](https://dbus.freedesktop.org/doc/dbus-tutorial.html#objects) still leaves unanswered the question - if interface is already prefixed then why add the same prefix to the object path? It just says "Namespacing object paths is smart, by starting them with the components of a domain name you own". O'rly? My guess it's that somebody just likes it that way. Then I would leave it up to the author here too.

org.sailfishos.usermanager.xml
monich: It's not good when such an obscure implementation detail finds its way into the interface. The whole purpose of an interface is to abstract out implementation details, isn't it?

sailfishusermanager.cpp
spiiroin: I vote "up to author" too.

mkenttala:
New version which no longer uses message arguments for the slots. Also now copies directories too and multiple other fixes here and there. Error handling is difficult when there are so many steps with the user handling, so the code pretty much still bails out at first sight of trouble.

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+<node>
+  <interface name="org.sailfishos.usermanager">
+    <method name="users">
+        <arg direction="out" type="a(ss)" name="users"/>
+        <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="QList&lt;SailfishUser&gt;"/>
+    </method>
+    <method name="addUser">
+        <arg direction="in" type="s" name="login_name"/>
```
Tomin1: I find these slightly confusing and not self-explanatory. Could they be username and (full_)name for example? Or maybe add some annotations at least?

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
...
+        <arg direction="in" type="s" name="user_name"/>
+    </method>
+    <signal name="userAdded">
```
Tomin1, line 13: This could also send id so that there is all the information about the user.

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
...
+        <arg type ="s" name="login_name"/>
+        <arg type ="s" name="user_name"/>
+    </signal>
+    <method name="removeUser">
+        <arg direction="in" type="s" name="login_name"/>
```
Tomin1: Is it better to use username or uid here? I can work with either though.

mkenttala: Username a bit easier and already implemented.

org.sailfishos.usermanager.xml
mkenttala: Switched to user_name and full_name

org.sailfishos.usermanager.xml
mkenttala: Done

sailfishusermanager.cpp
monich: matter of taste indeed ) I personally prefer stacking of ifs to returns from the middle - the former also forces separation of meaningful pieces of code into relatively small functions, because 4-5 nested ifs is already a bit too much for comfortable reading.

org.sailfishos.usermanager.xml
monich: uid is not used elsewhere in this API, and is easy enough to look up if necessary with `getpwnam`. If uid needs to be covered by this API too, then it needs to be added to `modifyUser` and `userModified` too.

sailfishusermanager.cpp
mkenttala: True, but I hate copy&pasteing those lu_free* stuff so decided to use this spaghetti approach instead, at least the code is indented :) Adding gotos is also something I refuse to do in C/C++.

Tomin1: I think separation of libuser parts and the interface could help with error handling too.

dbus-org.sailfishos.usermanager.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Sailfish login service
+
+[Service]
+Type=dbus
+BusName=org.sailfishos.usermanager
+ExecStart=/usr/bin/sailfish-logind
```
Tomin1: I got an error with this:
```
Mar 10 18:24:35 Sailfish systemd[1]: Starting Sailfish login service...
Mar 10 18:24:35 Sailfish systemd[7045]: dbus-org.sailfishos.usermanager.service: Failed at step EXEC spawning /usr/bin/sailfish-logind: No such file or directory
Mar 10 18:24:35 Sailfish systemd[1]: dbus-org.sailfishos.usermanager.service: Main process exited, code=exited, status=203/EXEC
Mar 10 18:24:35 Sailfish systemd[1]: Failed to start Sailfish login service.
Mar 10 18:24:35 Sailfish systemd[1]: dbus-org.sailfishos.usermanager.service: Unit entered failed state.
Mar 10 18:24:35 Sailfish systemd[1]: dbus-org.sailfishos.usermanager.service: Failed with result 'exit-code'.
```
Seems to be wrong binary name.

mkenttala: Fixed, along with some cleanups.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include <sys/stat.h>
+
+const char *DBUS_INTERFACE = "org.sailfishos.usermanager";
+const char *DBUS_OBJECT_PATH = "/";
+const char *USER_GROUP = "users";
+const char *GROUPS_IDS_FILE = "/usr/lib/setup/group.ids";
+const char *GROUPS_SYSTEM = "SYSTEM_GROUPS";
+const char *GROUPS_USER = "USER_GROUPS";
+const char *SKEL_DIR = "/etc/skel";
+const char *USER_HOME = "/home/%1";
```
Tomin1: You always create a QString out of this. Couldn't it just be a QStringLiteral here?
Applies to SKEL_DIR too, although it's implicitly turned into a QString.

libuserhelper.cpp
```diff
@@ -0,0 +1,234 @@
+#include "libuserhelper.h"
```
Tomin1, line 10: Missing copyright header, applies to libuserhelper.h too.

org.sailfishos.usermanager.xml
Tomin1: Well, true that it's easy to obtain. And the other way around using getpwuid is probably not more efficient either.

Something I started wondering now is that why this doesn't send SailfishUser instance? That's anyway what users returns. The other option is that it would only return username, I think.

sailfishusermanager.cpp
mkenttala: Fixed

libuserhelper.cpp
mkenttala: Fixed

org.sailfishos.usermanager.xml
mkenttala: Now sends the whole struct

sailfishusermanager.h
```diff
@@ -0,0 +1,73 @@
...
+#ifndef SAILFISHLOGIND_H
+#define SAILFISHLOGIND_H
+
+#include <QDBusContext>
+#include <QDBusArgument>
+
+class QTimer;
+class LibUserHelper;
+
+struct SailfishUser {
```
Tomin1: These things needed to use the D-Bus interface should probably be in some separate header file and available in devel subpackage. This struct, those operator methods below and definitions for D-Bus service names, interfaces and paths at least. And a pkgconf file too. And the xml of course.

libuserhelper.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+uint LibUserHelper::addGroup(const QString &group)
+{
+    struct lu_error *error = nullptr;
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, lu_prompt_console, NULL, &error);
+
+    if (!context) {
+        qWarning() << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return 0;
```
monich, line 27: Should this be `LU_VALUE_INVALID_ID` since it's obviously a failure? Same for `rv` in case if `lu_group_default()` fails.

mkenttala: I'd rather not add libuser dependencies to SailfishUserManager class now that they're all moved to separate helper class. That zero should be safe enough error value as it's reserved for root anyway.

monich: If zero represents an error, then `LU_VALUE_INVALID_ID` returned by `lu_ent_get_first_id()` needs to be replaced with zero, too. And if it returns zero.... then I don't know :) How about returning a positive signed int on success and `-1` on error?

mkenttala: Yep, missed that possible error situation. Now returns zero in that case.

sailfishusermanager.h
mkenttala: Devel package added.

org.sailfishos.usermanager.xml
Tomin1: I know I'm going forwards and backwards and whatever but... would it be better interface design if it just used usernames to identify users without these structs and then the client could check the information with getpwnam..?

rpm/user-managerd.spec
```diff
@@ -0,0 +1,49 @@
...
+
+%files
+%defattr(-,root,root,-)
+%{_bindir}/%{name}
+/lib/systemd/system/*.service
+%{_datadir}/dbus-1/system-services/*.service
+%{_sysconfdir}/dbus-1/system.d/*.conf
+
+%files devel
+%{_prefix}/include
```
Tomin1: This probably shouldn't contain /usr/include and /usr/lib/pkgconfig directories themselves.

sailfishusermanagerinterface.h
```diff
@@ -0,0 +1,43 @@
...
+ * BSD 3-Clause License, see LICENSE.
+ */
+
+#ifndef SAILFISHUSERMANAGERINTERFACE_H
+#define SAILFISHUSERMANAGERINTERFACE_H
+
+#include <QString>
+#include <QDBusArgument>
+
+#define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
```
Tomin1, line 16: QStringLiteral instead, please. The same on the line below.

sailfishusermanager.cpp
```diff
@@ -0,0 +1,246 @@
...
+#include <pwd.h>
+#include <grp.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+
+const char *USER_GROUP = "users";
+const char *GROUPS_USER = "USER_GROUPS";
+const auto GROUPS_IDS_FILE = QStringLiteral("/usr/lib/setup/group.ids");
+const auto GROUPS_SYSTEM = QStringLiteral("SYSTEM_GROUPS");
```
Tomin1: This is not used anywhere, right?

sailfishusermanagerinterface.h
inzanity: Just a note that if there ever are sub-interfaces, then the string literal is nicer, as they can be concatenated in the compiler.

pvuorela: Actually better avoid QStringLiteral or QString in headers, those would create separate instances by all the files including the header.

org.sailfishos.usermanager.xml
monich: Too bad that Unix password database doesn't have a clearly defined key. At any point of time both username and uid have to be unique, but either one can be changed and you can't rely on any of them as an immutable key.

but...

If we are to pick something as a key, that better be uid, I think. Do we support changing uid on the fly (including fixing file system permissions) in the first place? If not then let's use uid for identifying the users. As a bonus, it's smaller and comparing numbers is easier than comparing strings, no need for character set conversions and so on. Let's use numbers. Processors are better suited for handling numbers :)

Tomin1: No, we don't support changing uid and don't wish to implement that.

I don't know how glibc works internally regarding these getpwnam and getpwuid but text files are text so it needs convert the uids in the file to integers or alternatively convert the uid to a string and compare strings anyway if getpwuid is used. So I'm not sure it helps much here, although otherwise I agree. The internal representation on UI side uses uids anyway.

Also something I thought about, I wonder if getpw{name,uid} needs to go through the whole file every time or if stores some internal state to make it faster. There needs to be a few successive calls to them anyway if one wants to read all users. It probably doesn't matter much in practice as the files are not that big but it affects the time complexity still.

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+<node>
+  <interface name="org.sailfishos.usermanager">
+    <method name="users">
+        <arg direction="out" type="a(ssi)" name="users"/>
+        <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="QList&lt;SailfishUserManagerEntry&gt;"/>
+    </method>
+    <method name="addUser">
```
Tomin1, line 9: These methods don't have return values but would it make sense if they had? At least I see some benefit addUser returning uid or username (or user entry struct).

monich: Definitely needs to return uid, since username is provided as a parameter. Or both if the implementation is supposed to generate the username if you pass in an empty one.

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
...
+        <arg direction="in" type="s" name="user_name"/>
+        <arg direction="in" type="s" name="full_name"/>
+    </method>
+    <signal name="userAdded">
+        <arg type="(ssi)" name="user"/>
+        <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="SailfishUserManagerEntry"/>
+    </signal>
+    <method name="removeUser">
+        <arg direction="in" type="s" name="user_name"/>
```
monich: If uid is the (immutable) key in the user database, then the caller must pass uid here rather than username (which can change between the time when you looked it up and the time when this call reaches the service)

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
...
+    </method>
+    <signal name="userRemoved">
+        <arg type ="s" name="user_name"/>
```
monich: Again if uid is the key...

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
...
+    </signal>
+    <method name="modifyUser">
+        <arg direction="in" type="s" name="user_name"/>
```
monich: uid

org.sailfishos.usermanager.xml
```diff
@@ -0,0 +1,35 @@
...
+        <arg direction="in" type="s" name="new_user_name"/>
+        <arg direction="in" type="s" name="new_full_name"/>
+    </method>
+    <signal name="userModified">
+        <arg type ="s" name="user_name"/>
```
monich: uid (and I'm not sure if old username is needed)

org.sailfishos.usermanager.xml
mkenttala: Since libuser already has all these fields in it's entity strruct they may be sent to the  client as well, very little overhead. So at the moment I'm keeping this signal as it is.

sailfishusermanagerinterface.h
mkenttala: Used defines to avoid warnings from the compiler, didn't yet find a way to use QStringLiteral without complaints,

libuserhelper.cpp
```diff
@@ -0,0 +1,255 @@
...
+        lu_error_free(&error);
+        return false;
+    }
+
+    bool rv = true;
+    struct lu_ent *ent = lu_ent_new();
+    if (lu_user_lookup_id(context, uid, ent, &error)) {
+        lu_ent_set_string(ent, LU_USERNAME, new_user.toUtf8());
+        QString fixedName(new_name);
+        lu_ent_set_string(ent, LU_GECOS, fixedName.remove(',').toUtf8());
```
Tomin1: This should sanitize for ':' as well, unless libuser does that and we rely on that. Though, these checks could be on the API side as well.

libuserhelper.cpp
```diff
@@ -0,0 +1,255 @@
...
+    if (!context) {
+        qWarning() << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return false;
+    }
+
+    bool rv = true;
+    struct lu_ent *ent = lu_ent_new();
+    if (lu_user_lookup_id(context, uid, ent, &error)) {
+        lu_ent_set_string(ent, LU_USERNAME, new_user.toUtf8());
```
Tomin1: We should check username too. Or even better we should derive the username from full name and amend it if they collide with something existing. This might be better to do on the API side instead of this LibUserHelper.

sailfishusermanagerinterface.h
Tomin1: Well, pvuorela's comment is right that there could be more instances then. I did manage to use these inside QStringLiteral without problems on my code though. Maybe not too big of an issue then.

libuserhelper.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+    if (!context) {
+        qWarning() << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return 0;
+    }
+
+    uint rv = 0;
+    struct lu_ent *ent_group = lu_ent_new();
+    if (lu_group_default(context, group.toUtf8(), false, ent_group)) {
```
monich, line 32: Is UTF-8 really allowed as a group/user name or does it have to be limited to ASCII?

libuserhelper.cpp
mkenttala: Done

libuserhelper.cpp
mkenttala: New version generates clean user names.

sailfishusermanager.cpp
mkenttala: Let's add rollbacks later if testing finds them to be needed. Requires quite a lot of work for them to cover every possible situation.

sailfishusermanager.cpp
mkenttala: We'll figure out rollbacks later if needed.

libuserhelper.cpp
```diff
@@ -0,0 +1,266 @@
...
+    lu_ent_free(ent_group);
+    lu_end(context);
+
+    return rv;
+}
+
+bool LibUserHelper::addUserToGroup(const QString &user, const QString &group)
+{
+    struct lu_error *error = nullptr;
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, lu_prompt_console, NULL, &error);
```
inzanity: This `lu_prompt_console` looks scary in a daemon, which probably shouldn't even have a controlling terminal; is it possible to use NULL?

mkenttala: Nulled, didn't seem to print much debugging anyway.

libuserhelper.cpp
```diff
@@ -0,0 +1,266 @@
...
+    if (!context) {
+        qWarning() << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return false;
+    }
+
+    bool rv = true;
+    struct lu_ent *ent = lu_ent_new();
+
+    if (lu_user_lookup_name(context, new_user.toUtf8(), ent, &error)) {
```
Tomin1: This check seems to prevent keeping the same username. It would be good if you could modify the full name only, not touching username.

mkenttala: True, check removed as libuser makes the same check also.

Tomin1: Also with autogenerated usernames and no option to override it, it might be better if this didn't allow setting it either.

mkenttala: Yep, probably not needed so removed support for that.

libuserhelper.cpp
```diff
@@ -0,0 +1,272 @@
...
+    struct lu_ent *ent_group = lu_ent_new();
+    if (lu_group_default(context, group.toUtf8(), false, ent_group)) {
+        if (lu_group_add(context, ent_group, &error)) {
+            rv = lu_ent_get_first_id(ent_group, LU_GIDNUMBER);
+            if (rv == LU_VALUE_INVALID_ID) {
+                qWarning() << "Invalid group id";
+                rv = 0;
+            }
+        } else {
+            qWarning() << "Adding group failed:" << lu_strerror(error);
```
rainemak: qCWarning i.e. have logging category for this repo and component.

org.sailfishos.usermanagerd for instance

See e.g. https://github.com/sailfishos/sailfish-browser/blob/master/src/core/logging.cpp

libuserhelper.h
```diff
@@ -0,0 +1,28 @@
...
+class LibUserHelper
+{
+public:
+    LibUserHelper();
+    uint addGroup(const QString &group);
+    bool removeGroup(const QString &group);
+    bool addUserToGroup(const QString &user, const QString &group);
+    uint addUser(const QString &user, const QString &name, uint guid);
+    bool removeUser(uint uid);
+    bool modifyUser(uint uid, const QString &new_name);
```
rainemak: new_name -> newName

Qtified APIs should use camel-case

org.sailfishos.usermanager.conf
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE busconfig PUBLIC
+ "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+  "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+
+<busconfig>
+  <policy user="root">
+    <allow own="org.sailfishos.usermanager"/>
+    <allow send_destination="org.sailfishos.usermanager" send_interface="org.sailfishos.usermanager"/>
```
rainemak, line 8: Should we rename the repo to user-manager? Opinions?

libuserhelper.h
mkenttala: Renamed

libuserhelper.cpp
mkenttala: Logging category added

dbus-org.sailfishos.usermanager.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Sailfish login service
```
rainemak: Sailfish user manager service

libuserhelper.cpp
rainemak: Not seeing.

dbus-org.sailfishos.usermanager.service
mkenttala: Fixed

libuserhelper.cpp
mkenttala: Forgot to push to the server :)

rainemak:
LGTM

PR #2, opened 2020-03-23
[usermanager] Add access restrictions. Fixes JB#48598

Tomin1:
Deny access to operations on users below uid 100000. Allow users not in
sailfish-system group to access the daemon but restrict their access to
only themselves. Prevent removal of device owner.

### DIFF
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -23,6 +23,7 @@
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sailfishaccesscontrol.h>
 
 const char *USER_GROUP = "users";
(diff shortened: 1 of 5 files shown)


sailfishusermanager.cpp
```diff
@@ -162,6 +165,9 @@ bool SailfishUserManager::makeHome(const QString &user)
 
 uint SailfishUserManager::addUser(const QString &name)
 {
+    if (!hasAccessRights(SAILFISH_UNDEFINED_UID))
```
mkenttala, line 172: Either comment here or in the hasAccessRights method that SAILFISH_UNDEFINED_UI is a special case used to just check that the user/process has the required rights.

mkenttala:
LGTM

PR #3, opened 2020-03-25
[usermanager] User switch support. Fixes JB#47686

mkenttala:


sailfishusermanager.cpp
```diff
@@ -348,3 +349,65 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
 
     return true;
 }
+
+void SailfishUserManager::userSwitch(uint uid)
+{
+    if (!hasAccessRights(SAILFISH_UNDEFINED_UID)) {
+        auto message = QStringLiteral("Access denied");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::AccessDenied, message);
```
Tomin1: hasAccessRights already sends error replies.

sailfishusermanager.h
```diff
@@ -35,12 +35,14 @@ class SailfishUserManager : public QObject, protected QDBusContext
...
     void userRemoved(uint uid);
     void userModified(uint uid, const QString &new_name);
+    void userSwitched(uint uid);
 
 public slots:
     QList<SailfishUserManagerEntry> users();
     uint addUser(const QString &name);
     void removeUser(uint uid);
     void modifyUser(uint uid, const QString &new_name);
+    void userSwitch(uint uid);
```
Tomin1: switchUser would sound better IMO

sailfishusermanager.cpp
```diff
@@ -348,3 +349,65 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
 
     return true;
 }
+
+void SailfishUserManager::userSwitch(uint uid)
+{
+    if (!hasAccessRights(SAILFISH_UNDEFINED_UID)) {
+        auto message = QStringLiteral("Access denied");
```
monich: Hmm... What is `auto` in this case, any ideas? I suggest to avoid turning C++ code into a riddle.

sailfishusermanager.cpp
monich: Perhaps it should be called `checkAccess` or something like that. `hasAccessRights` indeed sounds like a check of access rights without any side effects.

sailfishusermanager.cpp
Tomin1: QString.

sailfishusermanager.cpp
Tomin1: Sounds like a good idea.

sailfishusermanager.cpp
```diff
@@ -348,3 +349,65 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+    uid_t currentUid;
+    if (sd_seat_get_active("seat0", nullptr, &currentUid) < 0) {
+        auto message = QStringLiteral("Failed to get current user id");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::Failed, message);
+        return;
+    }
+
+    QDBusInterface systemd(QStringLiteral("org.freedesktop.systemd1"), QStringLiteral("/org/freedesktop/systemd1"),
+                           QStringLiteral("org.freedesktop.systemd1.Manager"), QDBusConnection::systemBus());
```
monich: How about defining all string constants (especially those used repeatedly) somewhere in one place? As static objects, which might even reduce the size of the readonly-only data segment.

sailfishusermanager.cpp
```diff
@@ -348,3 +349,65 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+    reply = systemd.call(QStringLiteral("StartUnit"), QStringLiteral("autologin@%1.service").arg(uid),
+                         QStringLiteral("replace"));
+    if (!reply.isValid()) {
+        auto message = QStringLiteral("Failed to start autologin");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::Failed, message);
+
+        // Restart the old autologin
+        systemd.call(QStringLiteral("StartUnit"), QStringLiteral("autologin@%1.service").arg(currentUid),
+                     QStringLiteral("replace"));
```
monich: meh, copy/paste

sailfishusermanager.cpp
```diff
@@ -348,3 +349,65 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+        auto message = QStringLiteral("Failed to stop current user session");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::Failed, message);
+        return;
+    }
+
+    // Stop autologin service
+    reply = systemd.call(QStringLiteral("StopUnit"), QStringLiteral("autologin@%1.service").arg(currentUid),
+                         QStringLiteral("replace"));
+    if (!reply.isValid()) {
```
monich: This service is not serving D-Bus requests why this blocking call is pending, right? Could this cause any problems? Timeout is an obvious one, but could there be something trickier like a deadlock (A calls B, B calls C, C calls A - if all calls are blocking, it's a deadlock)?

sailfishusermanager.cpp
mkenttala: Works, gives no compiler warnings and this isn't called often so no performance hit.

sailfishusermanager.cpp
mkenttala: Done

sailfishusermanager.cpp
mkenttala: There's a default timeout for call, 25 secs according to docs.

sailfishusermanager.cpp
mkenttala: Consted

sailfishusermanager.h
mkenttala: Renamed

sailfishusermanager.cpp
```diff
@@ -297,13 +306,42 @@ bool SailfishUserManager::removeHome(uint uid)
...
+
 /* Check that calling D-Bus client is allowed to make the operation
  *
  * uid_to_modify is uid of the user that is going to be changed or removed.
  * Special value SAILFISH_UNDEFINED_UID can be used to denote non-existing user
  * that does not match to calling process' user but is in the valid range.
  */
-bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
+bool SailfishUserManager::checkAccessRights(uint uid_to_modify) {
     if (!calledFromDBus()) {
```
Tomin1: If you go with this approach that there are two checking methods, then I suggest that you move this to the checkCallerUid method. The point of this check is to make sure that the QDBusContext methods work properly.

The alternative could be to modify checkAccessRights so that it can work with switchUser too. Either redefine the use of SAILFISH_UNDEFINED_UID or add another value, I think. (Policy blocks addUser from other users anyway.) But either approach works.

sailfishusermanager.cpp
```diff
@@ -348,3 +370,54 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
 
     return true;
 }
+
+void SailfishUserManager::switchUser(uint uid)
```
Tomin1: I think this should probably also check that uid is an existing uid.

sailfishusermanager.cpp
monich: ok, riddle solved. but why does it have to be a riddle? oh well, I guess some people just like riddles. feel free to ignore my ranting

sailfishusermanager.cpp
Tomin1: It does sound to me that it's a bit bad idea to make these synchronous although it probably works alright in the end most of the time now. I'm not 100% sure that there won't be something calling this daemon while it is ending the session. Currently there isn't but I think that could change in later stages.

sailfishusermanager.cpp
Tomin1: I agree that it should be clear from the line of code what the return type is but I also think that here it's pretty clear, QStringLiteral is QString and no need to repeat that. You may disagree.

sailfishusermanager.cpp
mkenttala: Moved

sailfishusermanager.cpp
mkenttala: Check added

sailfishusermanager.cpp
mkenttala: Switched to async calls

sailfishusermanager.cpp
```diff
@@ -348,3 +370,98 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+        auto message = QStringLiteral("Failed to get current user id");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::Failed, message);
+        return;
+    }
+
+    m_switchUser = uid;
+    m_timer->start();
+
+    QDBusInterface systemd(SYSTEMD_MANAGER_DEST, SYSTEMD_MANAGER_PATH, SYSTEMD_MANAGER_INTERFACE, QDBusConnection::systemBus());
```
Tomin1: It feels a bit odd that this object is created separately for each call to the interface. It could be a member that is created when used first time. It would exist only if switchUser is used and would consume just a (null) pointer of memory otherwise.

mkenttala: Reusing interface now

monich: It definitely makes sense to make it global or something. Or even combine it with D-Bus call wrappers as a single entity responsible for D-Bus communication with systemd.

sailfishusermanager.cpp
```diff
@@ -34,11 +35,20 @@ const int HOME_MODE = 0700;
 const int QUIT_TIMEOUT = 60 * 1000; // One minute quit timeout
 const int MAX_RESERVED_UID = 99999;
 const int OWNER_USER_UID = 100000;
+const auto SYSTEMD_MANAGER_DEST = QStringLiteral("org.freedesktop.systemd1");
```
Tomin1: This is service name so maybe SYSTEMD_MANAGER_SERVICE instead.

sailfishusermanager.cpp
```diff
@@ -34,11 +35,20 @@ const int HOME_MODE = 0700;
...
+const auto SYSTEMD_MANAGER_STOP = QStringLiteral("StopUnit");
+const auto USER_SERVICE = QStringLiteral("user@%1.service");
+const auto SYSTEMD_MANAGER_REPLACE = QStringLiteral("replace");
+const auto AUTOLOGIN_SERVICE = QStringLiteral("autologin@%1.service");
 
 static_assert(SAILFISH_UNDEFINED_UID > MAX_RESERVED_UID,
               "SAILFISH_UNDEFINED_UID must be in the valid range of UIDs");
 
-SailfishUserManager::SailfishUserManager(QObject *parent) : QObject(parent), m_lu(new LibUserHelper())
+SailfishUserManager::SailfishUserManager(QObject *parent) : QObject(parent), m_lu(new LibUserHelper()), m_switchUser(0),
```
Tomin1: I think you should split this to multiple lines like we have in most projects. Also this doesn't seem to initialize m_currentUid and I don't see m_lu (LibUserHelper) being destructed.

sailfishusermanager.cpp
```diff
@@ -348,3 +371,97 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+
+    // Stop user service
+    QDBusPendingCall call = m_systemd->asyncCall(SYSTEMD_MANAGER_STOP, USER_SERVICE.arg(m_currentUid), SYSTEMD_MANAGER_REPLACE);
+    QDBusPendingCallWatcher *watcher = new QDBusPendingCallWatcher(call, this);
+    connect(watcher, &QDBusPendingCallWatcher::finished, this, &SailfishUserManager::userServiceStop);
+}
+
+void SailfishUserManager::userServiceStop(QDBusPendingCallWatcher *reply)
+{
+    if (!reply->isValid()) {
```
Tomin1: I found this really confusing initially. So this QDBusPendingCallWatcher implements indirection operator and points to  QDBusPendingReply which has isValid method. I think I prefer the style that is on [Qt documentation](https://doc.qt.io/archives/qt-5.6/qdbuspendingcallwatcher.html#details). It is used by most code anyway.

Also you should probably call deleteLater to destroy the watcher after it has served its purpose.

sailfishusermanager.cpp
```diff
@@ -348,3 +371,97 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+    connect(watcher, &QDBusPendingCallWatcher::finished, this, &SailfishUserManager::autologinServiceStart);
+}
+
+void SailfishUserManager::autologinServiceStart(QDBusPendingCallWatcher *reply)
+{
+    if (!reply->isValid()) {
+        auto message = QStringLiteral("Failed to start autologin");
+        qCWarning(lcSUM) << message;
+        emit userSwitchFailed(m_switchUser);
+        return;
```
Tomin1: Why return?

sailfishusermanager.cpp
```diff
@@ -348,3 +371,97 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+}
+
+void SailfishUserManager::autologinServiceStop(QDBusPendingCallWatcher *reply)
+{
+    if (!reply->isValid()) {
+        auto message = QStringLiteral("Failed to stop current user autologin");
+        qCWarning(lcSUM) << message;
+        emit userSwitchFailed(m_switchUser);
+        m_switchUser = 0;
+        return;
```
Tomin1: I wonder if this should do something to recover. Like reboot.

sailfishusermanager.cpp
monich: Call isError() (rather than isValid) and yes, deleteLater()

sailfishusermanager.cpp
```diff
@@ -348,3 +371,97 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+    // Stop user service
+    QDBusPendingCall call = m_systemd->asyncCall(SYSTEMD_MANAGER_STOP, USER_SERVICE.arg(m_currentUid), SYSTEMD_MANAGER_REPLACE);
+    QDBusPendingCallWatcher *watcher = new QDBusPendingCallWatcher(call, this);
+    connect(watcher, &QDBusPendingCallWatcher::finished, this, &SailfishUserManager::userServiceStop);
+}
+
+void SailfishUserManager::userServiceStop(QDBusPendingCallWatcher *reply)
+{
+    if (!reply->isValid()) {
+        auto message = QStringLiteral("Failed to stop current user session");
```
monich: This variable of a mysterious type is only used once, and can be dropped

sailfishusermanager.cpp
```diff
@@ -348,3 +371,97 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+        qCWarning(lcSUM) << message;
```
monich: I would include reply->error() in the error message

sailfishusermanager.cpp
```diff
@@ -348,3 +371,97 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+void SailfishUserManager::switchUser(uint uid)
+{
+    if (checkCallerUid() == SAILFISH_UNDEFINED_UID)
+        return;
+
+    if (m_switchUser) {
+        auto message = QStringLiteral("Already switching user");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::Failed, message);
+        return;
```
monich, line 396: What's D-Bus client supposed to do in this case? Retry? What if `m_switchUser` equals `uid` - that's not really an error, such (essentially, duplicate) request can be completed together with the one currently being processed.

I personally prefer to queue incoming requests and complete them one after another if they can't be done in parallel. That's simpler for the client (no need to implement retry logic) and better from the performance point of view (avoiding flood of retries). But of course that's slightly more complicated than handling one request at a time, so it may be too much to ask.

org.sailfishos.usermanager.xml
```diff
@@ -28,6 +28,15 @@
         <arg type ="u" name="uid"/>
         <arg type ="s" name="new_full_name"/>
     </signal>
+    <method name="switchUser">
```
monich: How to query the current user?

org.sailfishos.usermanager.xml
```diff
@@ -28,6 +28,15 @@
...
+        <arg direction="in" type="u" name="uid"/>
+    </method>
+    <signal name="userSwitched">
+        <arg type ="u" name="uid"/>
+    </signal>
+    <signal name="userSwitchFailed">
+        <arg type ="u" name="uid"/>
+    </signal>
   </interface>
```
monich, line 46: I would suggest a `currentUser` property (or `getCurrentUser` call if you prefer it that way), and rename `userSwitched` into `currentUserChanged`. And perhaps rename `switchUser` into `setCurrentUser` for consistency.

monich: There also should be a property that would tell clients that user is being switched, I think. That can be done in a number of ways, for example by introducing `nextUser` or `userSwitchingTo` property (or `getNextUser` call or something like that) together with `nextUserChanged` signal. Then if `currentUser != nextUser`, then user switch is in progress.

sailfishusermanager.cpp
monich: And if smart handling of incoming requests is too much to ask, then at least service should return a special error in this case (`Busy` or something), so that a smart client could insert a delay before retrying the request. And generally to let the client know that it's not really an error, the request can and should be retried.

sailfishusermanager.cpp
rainemak: :+1:

sailfishusermanager.cpp
mkenttala: Yep

sailfishusermanager.cpp
mkenttala: Ok

sailfishusermanager.cpp
mkenttala: Ok

sailfishusermanager.cpp
mkenttala: Ok

sailfishusermanager.cpp
mkenttala: Adding old service restarts to attempt recovery

sailfishusermanager.cpp
mkenttala: Introducing service specific errors, should be enough as there's only one client (settings UI) for this service.

org.sailfishos.usermanager.xml
mkenttala: Added currentUser method but keeping userSwithed as it better names it being response to userSwitch.

org.sailfishos.usermanager.xml
mkenttala: Adding currentUser method.

org.sailfishos.usermanager.xml
monich: I won't argue about naming, that's a very subjective topic. But don't think about it as a response. Response is sent directly to the caller and it's a separate type of D-Bus message. This is a signal, its recipients don't necessarily know and don't have to know what's caused it.

sailfishusermanager.cpp
```diff
@@ -348,3 +382,118 @@ bool SailfishUserManager::hasAccessRights(uint uid_to_modify) {
...
+    }
+
+    m_currentUid = currentUser();
+    if (m_currentUid == SAILFISH_UNDEFINED_UID)
+        return;
+
+    if (m_currentUid == uid) {
+        // No need to do anything
+        emit userSwitched(uid);
```
Tomin1: I would argue that this is now doing something (so it's not what the comment says). Usually it's best to not to emit if things don't change but it might be good to say that caller did something wrong instead. So basically spec the interface so that is shouldn't be called with current uid.

sailfishusermanagerinterface.h
```diff
@@ -12,10 +12,20 @@
 
 #include <QString>
 #include <QDBusArgument>
+#include <QDBusError>
 
 #define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
 #define SAILFISH_USERMANAGER_DBUS_OBJECT_PATH "/"
 
+#define SailfishUserManagerErrorBusy "org.sailfishos.usermanager.error.busy"
+#define SailfishUserManagerErrorHomeCreateFailed "org.sailfishos.usermanager.error.homecreatefailed"
```
Tomin1: I think errors should be TitleCased. Also Error is usually written with capital E on D-Bus interfaces.
https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-error

sailfishusermanager.cpp
mkenttala: Now sends an error here

sailfishusermanagerinterface.h
mkenttala: Titlecased

sailfishusermanagerinterface.h
```diff
@@ -12,10 +12,20 @@
 
 #include <QString>
 #include <QDBusArgument>
+#include <QDBusError>
 
 #define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
 #define SAILFISH_USERMANAGER_DBUS_OBJECT_PATH "/"
 
+#define SailfishUserManagerErrorBusy "org.sailfishos.usermanager.Error.Busy"
```
Tomin1, line 19: These are a bit inconsistent since some of these have word Error and some don't.

mkenttala: Yep

user-managerd.pro
```diff
@@ -15,7 +15,7 @@ DEFINES += QT_DEPRECATED_WARNINGS
 # You can also select to disable deprecated APIs only up to a certain version of Qt.
 #DEFINES += QT_DISABLE_DEPRECATED_BEFORE=0x060000    # disables all the APIs deprecated before Qt 6.0.0
 
-PKGCONFIG += libuser glib-2.0 sailfishaccesscontrol
+PKGCONFIG += libuser glib-2.0 sailfishaccesscontrol libsystemd-login mce-qt5
```
Tomin1: Needs matching build requires. Also I wonder if that libsystemd-login should be libsystemd or such. They did deprecate some of those names.

Tomin1: I wonder if that libsystemd-login was one of those deprecated libraries that were merged to the main library.

Tomin1: Apparently yes.
https://github.com/systemd/systemd/blob/master/NEWS#L7915
So this should probably be without -login suffix. Please check that it builds still if that is changed.

mkenttala: Removed. still seems to build.

mkenttala:
Renamed the methods as requested

org.sailfishos.usermanager.conf
```diff
@@ -22,5 +22,6 @@ Properties"/>
     <allow send_destination="org.sailfishos.usermanager" send_interface="org.sailfishos.usermanager" send_member="removeUser" />
     <allow send_destination="org.sailfishos.usermanager" send_interface="org.sailfishos.usermanager" send_member="modifyUser" />
     <allow send_destination="org.sailfishos.usermanager" send_interface="org.sailfishos.usermanager" send_member="users" />
+    <allow send_destination="org.sailfishos.usermanager" send_interface="org.sailfishos.usermanager" send_member="switchUser" />
```
Tomin1: This is still switchUser and currentUser is missing.

sailfishusermanagerinterface.h
```diff
@@ -12,10 +12,20 @@
 
 #include <QString>
 #include <QDBusArgument>
+#include <QDBusError>
```
Tomin1: This header doesn't actually use QDBusError for anything, does it?

Tomin1:
Almost good. One small thing and a nit.

org.sailfishos.usermanager.conf
mkenttala: Doh, too many places, fixed.

sailfishusermanagerinterface.h
mkenttala: Not anymore yes, removed.

Tomin1 (approved):

PR #4, opened 2020-03-31
[usermanager] Save user for the next boot. Fixes JB#49459

mkenttala:
Simple /etc/environment update so that the same user is used in the next boot too.


rainemak (approved):
Works nicely. LGTM

PR #5, opened 2020-04-01
[usermanager] Support for modifying users groups. Fixes JB#49418

mkenttala:


sailfishusermanager.h
```diff
@@ -49,6 +49,9 @@ public slots:
     void modifyUser(uint uid, const QString &new_name);
     void setCurrentUser(uint uid);
     uint currentUser();
+    QStringList usersGroups(uint uid);
+    void usersGroupsAdd(uint uid, const QString& group);
```
rainemak: Maybe just addGroup(uint uid, const QString &group), I'm assuming that this doesn't a thing i.e. succeeds if user alredy in the group.

For symmetry: removeGroup(uint uid, const QString& group)

Nit: ref should be on the group side :)

sailfishusermanager.cpp
```diff
@@ -535,3 +536,53 @@ void SailfishUserManager::updateEnvironment(uint uid)
...
+{
+    m_timer->start();
+    if (!checkAccessRights(uid))
+        return;
+
+    struct passwd *pwd = getpwuid(uid);
+    if (!pwd) {
+        auto message = QStringLiteral("User not found");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QDBusError::InvalidArgs, message);
```
rainemak: dedicated error message identifier strings similar to other strings. E.g. this could be SailfishUserManagerErrorUserNotFound

SailfishUserManagerErrorHomeCreateFailed...

sailfishusermanager.h
mkenttala: Renamed

sailfishusermanager.cpp
mkenttala: Added errors

Tomin1:
A question: would it be possible to make this such that it would change multiple groups at once? Best would be if that was somehow atomic, i.e. it would change all at once so it also fails all of them if it can't change them. If this is too much to ask, then that's fine.

mkenttala:
> A question: would it be possible to make this such that it would change multiple groups at once? Best would be if that was somehow atomic, i.e. it would change all at once so it also fails all of them if it can't change them. If this is too much to ask, then that's fine.

Getting complicated but I can give it a try, should be doable with arrays.

mkenttala:
> > A question: would it be possible to make this such that it would change multiple groups at once? Best would be if that was somehow atomic, i.e. it would change all at once so it also fails all of them if it can't change them. If this is too much to ask, then that's fine.
> 
> Getting complicated but I can give it a try, should be doable with arrays.

Libuser only supports editing groups one by one. So I can do this but have to enumerate through the groups and if something goes wrong have to revert somehow. Reverting without dropping user from a group it already belonged needs a list of those. This will take a while...

Tomin1:
> > > A question: would it be possible to make this such that it would change multiple groups at once? Best would be if that was somehow atomic, i.e. it would change all at once so it also fails all of them if it can't change them. If this is too much to ask, then that's fine.
> > 
> > 
> > Getting complicated but I can give it a try, should be doable with arrays.
> 
> Libuser only supports editing groups one by one. So I can do this but have to enumerate through the groups and if something goes wrong have to revert somehow. Reverting without dropping user from a group it already belonged needs a list of those. This will take a while...

Well, if it is too much maybe it's fine to edit them one by one. I'm thinking how it is used or might be used by the higher levels and there I think it makes sense to be able to change many of them at once. It's also possible that atomicity is not strictly required so it could just change those that it can do. Most likely it will succeed with all or none anyway I think.

If you end up doing it so that only some of the groups could be changed then it would be good to report the error so that it clearly says some groups were not added/removed instead of something that leads to assume that nothing was done. Most likely though it will succeed with all or none I think.

org.sailfishos.usermanager.xml
```diff
@@ -43,6 +43,19 @@
...
         <arg direction="out" type="u" name="uid"/>
     </method>
+    <method name="usersGroups">
+        <arg direction="in" type="u" name="uid"/>
+        <arg direction="out" type="a(s)" name="groups"/>
+        <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="QStringList"/>
+    </method>
+    <method name="addToGroup">
+        <arg direction="in" type="u" name="uid"/>
+        <arg direction="in" type="s" name="group"/>
```
monich: Why group name and not gid? For consistency with users, if nothing else. Besides, renaming of groups is easier to imagine than changing gids, which makes gid a better candidate for a group identifier. Slightly more efficient, too.

libuserhelper.cpp
```diff
@@ -110,6 +110,37 @@ bool LibUserHelper::addUserToGroup(const QString &user, const QString &group)
...
+        qCWarning(lcSUM) << "Error creating context";
+        return false;
+    }
+
+    bool rv = true;
+    struct lu_ent *ent = lu_ent_new();
+    if (lu_group_lookup_name(context, group.toLocal8Bit(), ent, &error)) {
+        GValue *value = g_slice_new0(GValue);
+        g_value_init(value, G_TYPE_STRING);
+        g_value_set_string(value, user.toUtf8());
```
monich, line 137: Are you sure this is safe?

`QString::toUtf()` returns `QByteArray` (which contains UTF-8 data converted from iternal UTF-16 representation). Then `QByteArray` gets cast to `const char*` which returns pointer to the internal buffer. That's the value which will be passed to `g_value_set_string()`. Now that compiler has the value it can destroy that temporary `QByteArray` which is no longer needed (along with the data which the pointer is pointing to). Are you sure that `QByteArray` is going to be destroyed **after** `g_value_set_string()` call and not before?

I'm quite sure I've seen similar cases when such a temporary `QByteArray` was being destroyed too early and the function received pointer to already deallocated memory. I would keep `QByteArray` on stack to prevent its buffer from being freed too early.

The same applies to `QString::toLocal8Bit()` - that one also returns `QByteArray`.

libuserhelper.cpp
```diff
@@ -110,6 +110,37 @@ bool LibUserHelper::addUserToGroup(const QString &user, const QString &group)
...
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, NULL, NULL, &error);
+    if (!context) {
+        qCWarning(lcSUM) << "Error creating context";
+        return false;
+    }
+
+    bool rv = true;
+    struct lu_ent *ent = lu_ent_new();
+    if (lu_group_lookup_name(context, group.toLocal8Bit(), ent, &error)) {
+        GValue *value = g_slice_new0(GValue);
```
monich, line 135: Who frees this `GValue` and how? Have you run this code under valgrind? If not, I suggest to do so.

libuserhelper.cpp
```diff
@@ -269,3 +300,33 @@ QString LibUserHelper::homeDir(uint uid)
...
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, NULL, NULL, &error);
+    if (!context) {
+        qCWarning(lcSUM) << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return QStringList();
+    }
+
+    QStringList rv;
+    struct lu_ent *ent = lu_ent_new();
+    if (lu_user_lookup_id(context, uid, ent, &error)) {
```
spiiroin, line 343: If this 'error' gets set, where is it supposed to get freed? Presumably in else block?

libuserhelper.cpp
```diff
@@ -269,3 +300,33 @@ QString LibUserHelper::homeDir(uint uid)
...
+        const char *user = lu_ent_get_first_string(ent, LU_USERNAME);
+        GValueArray *groups = lu_groups_enumerate_by_user(context, user, &error);
+        if (groups) {
+            for (uint i = 0; i < groups->n_values; i++) {
+                rv.append(g_value_get_string(g_value_array_get_nth(groups, i)));
+            }
+        } else {
+            qCWarning(lcSUM) << "Error getting users groups:" << lu_strerror(error);
+            lu_error_free(&error);
```
spiiroin, line 353: Side note: I took a peek at lu*error*() code. Looks "designed to be fragile" to me.

If error is set on entry to functions, library code will abort() - which is at least deterministic.

But then we have things like lu_groups_enumerate_by_user() which ensures error is not set when returning true, but AFAICT  *can* return false without setting error (when user==null) - in which case lu_strerror() here survives but lu_error_free() does null-deref crash and burn.

I have no idea how valid concern this might be.

sailfishusermanager.cpp
```diff
@@ -535,3 +536,53 @@ void SailfishUserManager::updateEnvironment(uint uid)
         file.close();
     }
 }
+
+QStringList SailfishUserManager::usersGroups(uint uid)
+{
+    m_timer->start();
```
spiiroin: nit: I guess this is "wait another minute before exiting due to inactivity" timer restart - imo it could use a bit more descriptive timer name / abstraction / stte.

org.sailfishos.usermanager.xml
mkenttala: I'm thinking UI would like to use strings here to enable/disable services for users, it does not know the gids but the names can be specified, some are already in use. Opinions @Tomin1 ?

libuserhelper.cpp
mkenttala: Specs say temporary object is valid through the full-expression, that should contain the whole line. Haven't seen any problems with this yet but lets keep in mind.

libuserhelper.cpp
mkenttala: Added frees.

libuserhelper.cpp
mkenttala: Added some error checks before freeing to avoid crashes. Indeed kibuser does not seem that robust.

libuserhelper.cpp
mkenttala: Added frees.

sailfishusermanager.cpp
mkenttala: Renamed

libuserhelper.cpp
monich: I prefer to keep `QByteArray` on stack in such cases.

org.sailfishos.usermanager.xml
monich: I wouldn't be thinking too much about UI at this point. UI can (and likely will) change in the future, D-Bus API is going to stay, and the cost of making a mistake is higher at lower level.

Tomin1: Well, this is actually quite interesting question. We have specified group names (see sailfish-setup) that we use but gids don't have any meaning really so from that point of view it's always an extra lookup to know how they match. Uids are much more meaningful IMO. Also we are not planning to allow creation or removal of groups (other than user's own group while creating or removing user) through this service and personally I don't think that's going to change in future either but who knows of course. I would say that for groups the names are fine but I'll leave that to you to decide. Settings plugin can adjust to either model as needed.

libuserhelper.cpp
spiiroin: @monich me too, but mostly because otherwise I have to think about how did it go again... The real problems I've seen/fixed have been of kind where the pointer is stored for later use, like if it were like:

    const char *text = user.toUtf8();
    g_value_set_string(value, text);

org.sailfishos.usermanager.xml
spiiroin: I'd say: user needs to be identified in a manner that is always resolvable in deterministic manner - and we are explicitly planning for cases like renaming user, or?.  The group bit here is just adjusting attributes of user which needs to hold water during the time of the change (and renaming group names without changing gid - what would be the use for that?). If the typical use case is adding groups by name, using gid would mean N x lookup code at client side vs. one place at service. I would prefer having potential hazard / complexity in one place.

monich: Yeah, cleanly renaming groups in our setup is just as impossible as changing numeric group ids, so I guess it doesn't really matter. When I have a choice, I usually prefer to use numbers rather than strings (because CPUs are much better suited for working with numbers, numbers have no issues with leading/trailing blank characters, upper/lower case and so on) but it's up to whoever is going to maintain this code. Which is not me :)

mkenttala: Using gids would be more robust but in this case the extra steps/work needed  from UI to use them is not worth it, keeping this as is.

sailfishusermanager.cpp
```diff
@@ -535,3 +536,80 @@ void SailfishUserManager::updateEnvironment(uint uid)
...
+
+QStringList SailfishUserManager::usersGroups(uint uid)
+{
+    m_exitTimer->start();
+    return m_lu->groups(uid);
+}
+
+void SailfishUserManager::addToGroups(uint uid, const QStringList &groups)
+{
+    m_exitTimer->start();
```
Tomin1: I wonder if this should be before or after checkAccessRights. At least it should be consistent so either change this or change the others.

sailfishusermanager.cpp
```diff
@@ -535,3 +536,80 @@ void SailfishUserManager::updateEnvironment(uint uid)
...
+    if (!checkAccessRights(uid))
```
Tomin1: User shouldn't be able to change their own groups. (Edit: well, this is actually prevented by D-Bus policy so having uid here instead of SAILFISH_UNDEFINED_UID doesn't really matter much. Edit2: And also this checks for uid being in range, so maybe this is good as is. Or we should make checkAccessRights arguments a bit more complicated. Or split it to more methods for different purposes and use those. There are options.)

Edit: Just ignore this. Or take it as a comment.

sailfishusermanager.cpp
mkenttala: Moved

Tomin1 (approved):
LGTM

libuserhelper.cpp
```diff
@@ -84,23 +87,64 @@ bool LibUserHelper::addUserToGroup(const QString &user, const QString &group)
...
         GValue *value = g_slice_new0(GValue);
         g_value_init(value, G_TYPE_STRING);
         g_value_set_string(value, user.toUtf8());
         lu_ent_add(ent, LU_MEMBERNAME, value);
         if (!lu_group_modify(context, ent, &error)) {
             qCWarning(lcSUM) << "Group modify failed:" << lu_strerror(error);
+            lu_error_free(&error);
             rv = false;
         }
+        g_slice_free(GValue, value);
```
spiiroin, line 108: The memory for value is released, but how about the content within?
Should there be g_value_unset(value) or something to undo resources allocated at g_value_set_string()?

libuserhelper.cpp
```diff
@@ -84,23 +87,64 @@ bool LibUserHelper::addUserToGroup(const QString &user, const QString &group)
...
+        GValue *value = g_slice_new0(GValue);
+        g_value_init(value, G_TYPE_STRING);
+        g_value_set_string(value, user.toUtf8());
+        lu_ent_del(ent, LU_MEMBERNAME, value);
+        if (!lu_group_modify(context, ent, &error)) {
+            qCWarning(lcSUM) << "Group modify failed:" << lu_strerror(error);
+            lu_error_free(&error);
+            rv = false;
+        }
+        g_slice_free(GValue, value);
```
spiiroin, line 145: the same, missing g_value_unset()?

spiiroin: and elsewhere too?

libuserhelper.cpp
monich: `GValue`s are not big (24 bytes on most platforms), they can be (and often are) allocated on stack. `g_value_unset()` is needed in any case, though.

mkenttala: Unset added

libuserhelper.cpp
mkenttala: Added

spiiroin:
LGTM

PR #6, opened 2020-04-20
[user-managerd] Delay user switching. Contributes to JB#49640

Tomin1:
Delay user switching until UI transition has ended. Don't let exit timer
to activate if user switching is in progress so that if switching takes
a long time it still works as intended.


mkenttala (approved):
LGTM

PR #7, opened 2020-04-22
[usermanager] Limit additional user count to 7. Contributes to JB#47647

mkenttala:


sailfishusermanager.cpp
```diff
@@ -211,6 +211,14 @@ uint SailfishUserManager::addUser(const QString &name)
         return 0;
     }
 
+    if (users().count() > SAILFISH_USERMANAGER_MAX_USERS) {
```
Tomin1: This does needlessly make calls to getpwnam so it would probably be better to call getgrnam directly and check how many users there are there. (Although, those getpwnam calls would be more useful if they also checked that the user actually exists but still it might have some effect to time complexity.)

And additionally this doesn't really need those SailfushUserManagerEntry structs but that's the least of overhead there is.

mkenttala: Fixed, also found the bug in users() method.

sailfishusermanager.cpp
```diff
@@ -211,6 +211,20 @@ uint SailfishUserManager::addUser(const QString &name)
         return 0;
     }
 
+    int count = 0;
+    struct group *grent = getgrnam(USER_GROUP);
+    if (grent)
```
Tomin1: Nit: it would be easier to read IMO if this outer check and that for loop inside both would have their own blocks, so basically:
```
    if (grent) {
        for (int i = 0; grent->gr_mem[i]; i++) {
            if (getpwnam(grent->gr_mem[i]))
                count++;
        }
    }
```

Tomin1 (approved):
LGTM. Just one minor nit that you may want to change.

sailfishusermanager.cpp
```diff
@@ -211,6 +211,22 @@ uint SailfishUserManager::addUser(const QString &name)
...
+    int count = 0;
+    struct group *grent = getgrnam(USER_GROUP);
+    if (grent) {
+        for (int i = 0; grent->gr_mem[i]; i++) {
+            if (getpwnam(grent->gr_mem[i]))
+                count++;
+        }
+    }
+    if (count > SAILFISH_USERMANAGER_MAX_USERS) {
+        // Just > above because master user is also member of users group
```
spiiroin, line 223: AFAIK The number of users we can have is 7 and that includes the master user. So, already having 7 users and still proceeding to add 8th should be wrong - or am I missing something?

mkenttala: Luks has eight slots where one is reserved for master or are there more reservations?

Tomin1: One of them is reserved for backup (temporary). So 7 users including master...

mkenttala: Ok, fixing this in the pc version fix commit.

spiiroin: Sorry to flog a dead horse / already merged pr, but AFAICT it should be "if (count >= 7)" not "if (count > 7)". We have 8 luks slots, 7 of those available for 7 users - one of which happens to be the master. Which yields SAILFISH_USERMANAGER_MAX_USERS = 7 -> counting already present users, finding out that we have 7 of them, and still proceeding to create 8th one = wrong / I am missing something fundamental here.

spiiroin: oh well, it was fixed in not-closed pr ;-)

PR #8, opened 2020-04-24
[usermanager] Proper pkgconfig version. Contributes to JB#47647

mkenttala:
Didn't find a way to make this qt create_pc easily, would have needed to move everything to subdirs.

### DIFF
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -219,8 +219,8 @@ uint SailfishUserManager::addUser(const QString &name)
                 count++;
         }
     }
-    if (count > SAILFISH_USERMANAGER_MAX_USERS) {
-        // Just > above because master user is also member of users group
+    if (count > (SAILFISH_USERMANAGER_MAX_USERS - 1)) {
+        // Master user reserves one slot above
         auto message = QStringLiteral("Maximum number of users reached");
         qCWarning(lcSUM) << message;
         sendErrorReply(QStringLiteral(SailfishUserManagerErrorMaxUsersReached), message);
--- a/sailfishusermanagerinterface.h
+++ b/sailfishusermanagerinterface.h
@@ -15,6 +15,8 @@
 
 #define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
 #define SAILFISH_USERMANAGER_DBUS_OBJECT_PATH "/"
+
+// Luks has eight slots where one slot is reserved for backup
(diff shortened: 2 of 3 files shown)


sailfishusermanagerinterface.h
```diff
@@ -15,7 +15,7 @@
 
 #define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
 #define SAILFISH_USERMANAGER_DBUS_OBJECT_PATH "/"
-#define SAILFISH_USERMANAGER_MAX_USERS 7
+#define SAILFISH_USERMANAGER_MAX_USERS 6
```
Tomin1: I would prefer to include master user. Or is there some compelling reason to count only additional users?

mkenttala: I was thinking this in more of naming thing, user != master.  If objections I can revert this and drop the count in addUser too.

Tomin1: Hmm. A call to users() returns also master user, right? That's why I'd think this count should also include master user. But I mean as long as it's clear and documented I'm okay with it. Currently I think it's quite confusing.

mkenttala: Fixed

sailfishusermanager.cpp
```diff
@@ -219,8 +219,8 @@ uint SailfishUserManager::addUser(const QString &name)
                 count++;
         }
     }
-    if (count > SAILFISH_USERMANAGER_MAX_USERS) {
-        // Just > above because master user is also member of users group
+    if (count > (SAILFISH_USERMANAGER_MAX_USERS - 1)) {
+        // Luks has 8 slots where one is reserved for backup and one for master
```
Tomin1: This comment about number of slots feels a bit out of place. Wouldn't it be better to locate that information next to the constant definition?

sailfishusermanager.cpp
```diff
@@ -219,8 +219,8 @@ uint SailfishUserManager::addUser(const QString &name)
                 count++;
         }
     }
-    if (count > SAILFISH_USERMANAGER_MAX_USERS) {
-        // Just > above because master user is also member of users group
+    if (count > (SAILFISH_USERMANAGER_MAX_USERS - 1)) {
```
Tomin1, line 222: Hmm. I can see that you don't consider master as an user here but if it was considered as an user then this could be `count >= SAILFISH_USERMANAGER_MAX_USERS`.

spiiroin:
:+1:

Tomin1 (approved):

PR #9, opened 2020-04-29
[user-mangerd] Add proper systemd job control. Fixes JB#49717

Tomin1:
This should clear up any issues with job management which improves user
switching a lot. It also somewhat robustifies error handling in case
some of the systemd units end with failure exit code.

Add SystemdManager class that is handles starting and stopping units. It
reports that job has finished only after the job has been removed. It
also improves code reuse as all units are handled with the same code.

### DIFF
--- a/systemdmanager.cpp
+++ b/systemdmanager.cpp
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2020 Open Mobile Platform LLC.
+ *
+ * All rights reserved.
+ *
+ * BSD 3-Clause License, see LICENSE.
+ */
+
+#include "systemdmanager.h"
+#include "logging.h"
+#include <QDBusInterface>
+#include <QDBusObjectPath>
+#include <QDBusPendingCall>
+#include <QDBusPendingReply>
+
+namespace Systemd {
+const auto Service = QStringLiteral("org.freedesktop.systemd1");
+const auto ManagerPath = QStringLiteral("/org/freedesktop/systemd1");
+const auto ManagerInterface = QStringLiteral("org.freedesktop.systemd1.Manager");
+const auto Replace = QStringLiteral("replace");
+const auto StartUnit = QStringLiteral("StartUnit");
+const auto StopUnit = QStringLiteral("StopUnit");
+const auto ResultDone = QStringLiteral("done");
+const auto ResultSkipped = QStringLiteral("skipped");
+}
+
+// This is currently implemented so that it can do one thing at a time
+// as there is no need for anything more complicated.
+// It would be quite easily possible to add individual queues with
+// queue numbers if needed.
+
+SystemdManager::SystemdManager(QObject *parent) :
+    QObject(parent),
+    m_systemd(new QDBusInterface(Systemd::Service, Systemd::ManagerPath,
+                                 Systemd::ManagerInterface,
+                                 QDBusConnection::systemBus(), this))
+{
+    if (!m_systemd->isValid())
+        qCCritical(lcSUM) << "Could not create interface to systemd, can not function!";
+    if (!connect(m_systemd, SIGNAL(JobRemoved(uint, QDBusObjectPath, QString, QString)),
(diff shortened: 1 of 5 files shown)


systemdmanager.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (m_pendingCall || !m_currentJob.isEmpty())
+        return;
+
+    qCDebug(lcSUM) << "Process next systemd job";
+
+    QDBusPendingCall call = m_systemd->asyncCall(
+            (m_jobs.first().type == StopJob) ? Systemd::StopUnit : Systemd::StartUnit,
+            m_jobs.first().unit, Systemd::Replace);
+    QDBusPendingCallWatcher *m_pendingCall = new QDBusPendingCallWatcher(call, this);
+    connect(m_pendingCall, &QDBusPendingCallWatcher::finished, this, &SystemdManager::onJobStarted);
```
monich: `SystemdManager::onJobStarted` is invoked even if job type was `StopJob`, right?

Tomin1: Job != Unit. In systemd speak job is a task of systemd and thus starting or stopping a unit is a job. So yes, it's correct to invoke that on both types of job.

monich: Yeah, I solved that riddle. Nice way to confuse the reader!

mkenttala:
LGTM, nothing bad catches my eye.

systemdmanager.cpp
monich: But it's basically a queue of D-Bus calls, and a "job" is a D-Bus call, right? It's just being referred to as a "job". To confuse the reader :)

Tomin1: I'm not sure if I follow. So for each (systemd) job we have a D-Bus call to start it and then a signal that tell that the job was removed (i.e. has ended to a failure or success). We need to handle both of these.

monich: If onJobStarted was onCallFinished , and m_jobs was m_callQueue or something like this, the code would be significantly easier to understand. At least to me.

When I saw "job" I was thinking that it was another name for a system unit. But in the end it turned out to be another name for a D-Bus call. That was confusing.

Basically, I suggest to refer to D-Bus calls as "D-Bus calls", that's all. Unless the goal is to confuse the reader. In that case - good job, you have confused me!

Tomin1: Well of course my goal is to confuse you. :roll_eyes: (For the record, it's not.) No need to be so hostile about this. I think you are right that I could probably name onJobStarted better. Still I think these are not D-Bus calls as such but instead they are systemd jobs and involve more than just a D-Bus call. Note that I'm only removing them from queue once the job is done. But I can still reconsider the naming and add some explanatory comments if they seem useful.

monich: oh come on, "good job" was a calambur, no hostility intended or anything. There were systemd-style job control systems on other OSes where units were called jobs, that experience of mine contributed to confusion.

monich: In other words, feel free to ignore.

Tomin1:
I noticed that updateEnvironment was being called with uid = 0, fixed that. Also changed the onJobStarted to pendingCallFinished as there is only one kind of pending call and that kind of naming is inline with some other parts of the code too. And also added explanatory naming why the struct is called "Job" and thus it's used in other places as well. I don't really think we should change that as systemd uses that naming and this is SystemdManager class.

Edit: Pushed again, this time just changed one word in commit message. Apparently systemd language caught me too (there was word job when it should have been unit really).

Edit again: Pushed once more. I noticed that I had forgot to parent SystemdManager when I created it. (Memory management is easy to get wrong...) Then when testing that I noticed that I got that comparison to uid wrong way around (converted it from assert to return earlier). If you think it should be an assert, please comment.

Edit: And the latest change was just to change that "Contributes to" to "Fixes".

Edit: ... checked Q_ASSERT_X arguments, they were also in wrong order... Pushed that change too.

mkenttala (approved):
Still LGTM

PR #11, opened 2020-05-12
[user-managerd] Enable lockcode after user change. Fixes JB#49623

mkenttala:


spiiroin:
The change itself seems fine.

However, seeing that this locks device when user switch initiated by user manager has successfully completed and we have other services depending on signaling from the same trigger... Do we need to worry about something switching user without involving user manager? Probably that can happen, but not without root privileges - or?

Tomin1:
I would think that it would be better if device lock was the one that decided if it should lock. I think it could detect if session on seat0 was changed. Session identifiers are unique and and there are ways to watch for those changes ([a lot like how it's used here](https://git.sailfishos.org/mer-core/nemo-qml-plugin-systemsettings/blob/c6e14bfd875dde3cbf0fd106273ea529ede7304b/src/userinfo.cpp#L548)).

mkenttala:
Yes, a bit more elegant way to do this in the lock daemon. Discarding this one.

PR #15, opened 2020-05-28
[usermanager] Remove user environment files when user removed. Fixes …

mkenttala:
Don't know if that shadow-utils userdel scripting is really that much needed but it's there now.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -14,6 +14,8 @@
 int main(int argc, char *argv[])
 {
     QCoreApplication app(argc, argv);
+    if (argc > 2 && !strcmp(argv[1], "--removeUserFiles"))
+        return SailfishUserManager::removeUserFiles(argv[2]);
     SailfishUserManager daemon;
     return app.exec();
 }
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -44,6 +44,7 @@ const auto AUTOLOGIN_SERVICE = QStringLiteral("autologin@%1.service");
 const auto ENVIRONMENT_FILE = QStringLiteral("/etc/environment");
 const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
+const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
 
 static_assert(SAILFISH_UNDEFINED_UID > MAX_RESERVED_UID,
               "SAILFISH_UNDEFINED_UID must be in the valid range of UIDs");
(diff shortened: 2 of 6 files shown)


main.cpp
```diff
@@ -14,6 +14,8 @@
 int main(int argc, char *argv[])
 {
     QCoreApplication app(argc, argv);
+    if (argc > 2 && !strcmp(argv[1], "--removeUserFiles"))
```
spiiroin, line 17: I'd much prefer seeing some sort of more or less conforming arg parser in here ;-)

sailfishusermanager.cpp
```diff
@@ -287,6 +288,18 @@ uint SailfishUserManager::addUser(const QString &name)
...
+int SailfishUserManager::removeUserFiles(const char *user)
+{
+     struct passwd *pwd = getpwnam(user);
+     if (pwd) {
+         QDir dir(USER_ENVIRONMENT_DIR.arg(pwd->pw_uid));
+         if (dir.removeRecursively())
+             return 0;
+         qCWarning(lcSUM) << "Removing user environment directory failed";
+     }
+     return 1;
```
spiiroin: Logic wise this would look like a "bool function()", but as it is actually exit code -> using EXIT_SUCCESS / EXIT_FAILURE would be cleared than literal 0 / 1.

sailfishusermanager.cpp
```diff
@@ -287,6 +288,18 @@ uint SailfishUserManager::addUser(const QString &name)
     return uid;
 }
 
+int SailfishUserManager::removeUserFiles(const char *user)
+{
+     struct passwd *pwd = getpwnam(user);
+     if (pwd) {
+         QDir dir(USER_ENVIRONMENT_DIR.arg(pwd->pw_uid));
+         if (dir.removeRecursively())
```
spiiroin: Should there be some check before going nucular - e.g. is clearing files for master user also ok?

sailfishusermanager.cpp
```diff
@@ -304,6 +317,10 @@ void SailfishUserManager::removeUser(uint uid)
         return;
     }
 
+    struct passwd *pwd = getpwuid(uid);
+    if (pwd)
+        removeUserFiles(pwd->pw_name);
```
spiiroin: uid -> name via getpwuid(), then that returned name -> uid via getpwnam() = potentially disastrous pitfall given the context of recursive directory wiping

"... return value may point to a static area, and may be overwritten by subsequent calls to getpwent(3), getpwnam(), or getpwuid()."

spiiroin (changes requested):
That chained getpwuid() getpwnam() should be fixed. The rest are nits.

spiiroin (approved):

main.cpp
mkenttala: Sure needed if there ever comes more arguments to parse. Not compilicating things for now as there in only this one simple case.

sailfishusermanager.cpp
mkenttala: If dbus call was used this was already checked and denied but if ran directly or through userdel then the user is already lost.

PR #16, opened 2020-06-05
[usermanager] Add support for user removal scripts. Fixes JB#48821

mkenttala:
…JB#48821

### DIFF
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -45,6 +45,7 @@ const auto ENVIRONMENT_FILE = QStringLiteral("/etc/environment");
 const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
+const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/remove.d");
 
 static_assert(SAILFISH_UNDEFINED_UID > MAX_RESERVED_UID,
               "SAILFISH_UNDEFINED_UID must be in the valid range of UIDs");
@@ -135,7 +136,7 @@ bool SailfishUserManager::addUserToGroups(const QString &user)
     while (!file.atEnd()) {
         line = file.readLine();
         if (line.startsWith(GROUPS_USER)) {
-            foreach (const QString& group, line.mid(strlen(GROUPS_USER) + 1).split(',')) {
+            for (const QString& group : line.mid(strlen(GROUPS_USER) + 1).split(',')) {
                 if (!m_lu->addUserToGroup(user, group.trimmed())) {
                     file.close();
                     return false;
@@ -160,12 +161,12 @@ bool SailfishUserManager::copyDir(const QString &source, const QString &destinat
         return false;
     }
 
-    foreach (const QString &dir, sourceDir.entryList(QDir::Dirs | QDir::NoDotAndDotDot | QDir::Hidden)) {
+    for (const QString &dir : sourceDir.entryList(QDir::Dirs | QDir::NoDotAndDotDot | QDir::Hidden)) {
         if (!copyDir(sourceDir.path() + '/' + dir, destination + '/' + dir, uid, guid))
             return false;
     }
 
-    foreach (const QString &file, sourceDir.entryList(QDir::Files | QDir::NoDotAndDotDot | QDir::Hidden)) {
+    for (const QString &file : sourceDir.entryList(QDir::Files | QDir::NoDotAndDotDot | QDir::Hidden)) {
         QString destFile = QString("%1/%2").arg(destination).arg(file);
         if (!QFile::copy(QString("%1/%2").arg(sourceDir.path()).arg(file), destFile)) {
             qCWarning(lcSUM) << "Failed to copy file";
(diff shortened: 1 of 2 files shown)


sailfishusermanager.cpp
```diff
@@ -45,6 +45,7 @@ const auto ENVIRONMENT_FILE = QStringLiteral("/etc/environment");
 const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
+const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-manager/removal");
```
Tomin1: How about `/usr/share/user-managerd/remove.d`? I'm just thinking that we use names like `user-managerd` (the repo) and `usermanager` (e.g. D-Bus service) but not `user-manager` so I would use either of those. Name of the repo feels more natural to me here but either works. And that `removal` felt a bit out of place so since the interface is `removeUser()` maybe we should mirror that so either `remove` or `removeUser`. I added `.d` to the end because it feels like convention to name those directories that way.

sailfishusermanager.cpp
```diff
@@ -290,11 +291,20 @@ uint SailfishUserManager::addUser(const QString &name)
...
-    qCWarning(lcSUM) << "Removing user environment directory failed";
-    return EXIT_FAILURE;
+        rv = EXIT_SUCCESS;
+    else
+        qCWarning(lcSUM) << "Removing user environment directory failed";
+
+    // Execute user removal scripts
+    QDir removal(USER_REMOVE_SCRIPT_DIR);
+    foreach (const QString &entry, removal.entryList(QDir::Files | QDir::NoDotAndDotDot)) {
+        QProcess::execute(USER_REMOVE_SCRIPT_DIR + '/' + entry, QStringList() << QString::number(uid));
```
Tomin1: I think this should check if the process succeeded and if not, then print that to log. Maybe with command output if that's not already printed there (I didn't check!). It doesn't make sense to stop if a script fails but it should be clear and visible.

sailfishusermanager.cpp
```diff
@@ -290,11 +291,20 @@ uint SailfishUserManager::addUser(const QString &name)
...
-        return EXIT_SUCCESS;
-    qCWarning(lcSUM) << "Removing user environment directory failed";
-    return EXIT_FAILURE;
+        rv = EXIT_SUCCESS;
+    else
+        qCWarning(lcSUM) << "Removing user environment directory failed";
+
+    // Execute user removal scripts
+    QDir removal(USER_REMOVE_SCRIPT_DIR);
+    foreach (const QString &entry, removal.entryList(QDir::Files | QDir::NoDotAndDotDot)) {
```
Tomin1: Does this actually need `NoDotAndDotDot` when it has `Files`? Then again it doesn't really hurt.

sailfishusermanager.cpp
Tomin1: Oh and also I think this could skip files without execution permissions. Could be handy if you can skip some of them that way. Just an idea, technically it probably works already as it shouldn't execute those anyway.

sailfishusermanager.cpp
```diff
@@ -290,11 +291,21 @@ uint SailfishUserManager::addUser(const QString &name)
...
-    return EXIT_FAILURE;
+        rv = EXIT_SUCCESS;
+    else
+        qCWarning(lcSUM) << "Removing user environment directory failed";
+
+    // Execute user removal scripts
+    QDir removal(USER_REMOVE_SCRIPT_DIR);
+    foreach (const QString &entry, removal.entryList(QDir::Files | QDir::Executable)) {
+        if (QProcess::execute(USER_REMOVE_SCRIPT_DIR + '/' + entry, QStringList() << QString::number(uid)))
+            qCWarning(lcSUM) << "User remove script" << USER_REMOVE_SCRIPT_DIR + '/' + entry << "failed";
```
Tomin1: It could be useful to ~~return~~ print the exit code too.

Tomin1: Sorry, not return but print of course!

rainemak: Nit pick: prefer C++ for range over the foreach => also Qt code has been moving away from foreach...

```
for (const QString &entry : removal.entryList(QDir::Files | QDir::Executable)) {
```

Tomin1 (approved):
LGTM. You could change those other foreaches to new style C++ for loops if you want but this is good like this too.

sailfishusermanager.cpp
```diff
@@ -290,11 +291,22 @@ uint SailfishUserManager::addUser(const QString &name)
...
-    qCWarning(lcSUM) << "Removing user environment directory failed";
-    return EXIT_FAILURE;
+        rv = EXIT_SUCCESS;
+    else
+        qCWarning(lcSUM) << "Removing user environment directory failed";
+
+    // Execute user removal scripts
+    QDir removal(USER_REMOVE_SCRIPT_DIR);
+    for (const QString &entry : removal.entryList(QDir::Files | QDir::Executable)) {
+        int exitCode = QProcess::execute(USER_REMOVE_SCRIPT_DIR + '/' + entry, QStringList() << QString::number(uid));
```
spiiroin, line 304: Too late etc. More than once bitten: executing everything that exists opens door for weird mishaps via having backup files / package upgrade artefacts getting executed too... enforcing some sort of naming convention (regexp/whatnot) would avoid such things.

spiiroin: Also having deterministic execute order never hurts, now it is more or less random due to implicit NoSort.

Tomin1: > Too late

Well, I would argue it's too late only once the software has been released (or so, there can be other deadlines as well) so this should be about as easy as creating another PR with the fixes.

PR #17, opened 2020-06-05
[user-managerd] Start user@.service explicitly. Fixes JB#49774

Tomin1:
Start user@.service explicitly when starting new session. Previously this relied on autologin@.service to start it through pam_systemd but in practice it may skip that if user-.slice is still active and thus user
session doesn't start properly. It shouldn't matter that pam_systemd is also starting user@.service because then systemd just reports the start job as 'done' and continues. Use 'fail' instead of 'replace' to not make changes to starting process if pam_systemd was faster.


mkenttala (approved):
LGTM

PR #18, opened 2020-06-11
[user-managerd] Add introspection to default context policy. Fixes JB#50203

Tomin1:
Add introspection to default context policy. Also remove extra linebreaks and add some spaces.


mkenttala (approved):
Thumbed

PR #20, opened 2020-06-23
[user-managerd] Implement user quota limits. Fixes JB#47805

Tomin1:
Quota limits are set on user creation. This implements fixed limits
without configurability. Only block limits are defined, inode limits are
cleared if they happen to be set for the uid.

### DIFF
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -19,14 +19,20 @@
 #include <QDir>
 #include <QString>
 
-#include <pwd.h>
+#include <errno.h>
 #include <grp.h>
-#include <unistd.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <sailfishaccesscontrol.h>
-#include <systemd/sd-login.h>
+#include <pwd.h>
 #include <qmcecallstate.h>
+#include <sailfishaccesscontrol.h>
+#include <sys/mount.h>
+#include <sys/quota.h>
+#include <sys/stat.h>
+#include <sys/statvfs.h>
+#include <sys/types.h>
+#include <systemd/sd-login.h>
+#include <unistd.h>
+
+namespace {
 
 const char *USER_GROUP = "users";
 const char *GROUPS_USER = "USER_GROUPS";
@@ -46,9 +52,18 @@ const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
 const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/remove.d");
+const quint64 MAXIMUM_QUOTA_LIMIT = 2000000000ULL;
 
 static_assert(SAILFISH_UNDEFINED_UID > MAX_RESERVED_UID,
               "SAILFISH_UNDEFINED_UID must be in the valid range of UIDs");
+
+QByteArray findHomeDevice()
+{
+    QStorageInfo info(USER_HOME.arg(""));
+    return info.device();
+}
+
+};
 
 SailfishUserManager::SailfishUserManager(QObject *parent) :
     QObject(parent),
@@ -281,6 +296,8 @@ uint SailfishUserManager::addUser(const QString &name)
         return 0;
     }
 
+    setUserLimits(uid);
+
     SailfishUserManagerEntry entry;
     entry.user = user;
     entry.name = name;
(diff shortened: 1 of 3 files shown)


sailfishusermanager.cpp
```diff
@@ -46,10 +52,19 @@ const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
 const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/remove.d");
+const quint64 MAXIMUM_QUOTA_LIMIT = 2000000000LL;
```
spiiroin: nit: LL suffix = signed -> ULL would be more appropriate for quint64.

sailfishusermanager.cpp
```diff
@@ -309,6 +326,50 @@ int SailfishUserManager::removeUserFiles(uint uid)
...
 }
 
+/*
+ * Sets user quota limits, if supported by kernel and enabled on /home filesystem
+ */
+void SailfishUserManager::setUserLimits(uint uid)
+{
+    struct statvfs info;
+    errno = 0;
+    if (statvfs("/home", &info) < 0) {
```
spiiroin: once bitten etc... Suitable libc/kernel mismatch can lead to statvfs() and like filling in only part of info buffer, leaving the rest to whatever happened to be in stack and return success. This can then lead to really confusing and hard to figure out situations -> I've made it a habit to clear such buffers before use.

sailfishusermanager.cpp
```diff
@@ -309,6 +326,50 @@ int SailfishUserManager::removeUserFiles(uint uid)
...
+        if (softlimit > MAXIMUM_QUOTA_LIMIT)
+            softlimit = MAXIMUM_QUOTA_LIMIT;
+        // Hard limit is 120 % of soft limit
+        quint64 hardlimit = softlimit * 120 / 100;
+        qCDebug(lcSUM) << "Setting quota limits for" << uid << "to"
+                       << hardlimit << "and" << softlimit << "bytes";
+        // Sets block limits and clears inode limits
+        struct if_dqblk quota = {
+            .dqb_bhardlimit = (fsblkcnt_t)(hardlimit / BLOCK_SIZE),
+            .dqb_bsoftlimit = (fsblkcnt_t)(softlimit / BLOCK_SIZE),
```
spiiroin: I'm wondering: what if info.f_bsize != BLOCK_SIZE?

If (queried) device capacity and (set) limits for device both deal with blocks, would it be more correct / less likely to misfire if blocks -> bytes -> blocks roundtrip were not made at all?

sailfishusermanager.cpp
```diff
@@ -309,6 +326,50 @@ int SailfishUserManager::removeUserFiles(uint uid)
...
+            .dqb_curspace = 0,
+            .dqb_ihardlimit = 0,
+            .dqb_isoftlimit = 0,
+            .dqb_curinodes = 0,
+            .dqb_btime = 0,
+            .dqb_itime = 0,
+            .dqb_valid = QIF_LIMITS
+        };
+        errno = 0;
+        if (quotactl(QCMD(Q_SETQUOTA, USRQUOTA), findHomeDevice().data(), (uid_t)uid, (caddr_t)&quota) < 0) {
```
spiiroin, line 361: query hard coded "/home" vs set "/home/" obtained in roundabout way?

spiiroin (changes requested):

sailfishusermanager.cpp
Tomin1: Actually the mistake here is to think that filesystem is defined in f_block sized units but the man page actually says its f_frsize that it uses. Quota uses blocks of BLOCK_SIZE which is why there is the round trip. I guess if those two always happen to be of the same size, then conversion to bytes could be avoided. f_bsize was 4096 on the device I used and BLOCK_SIZE was 1024.

spiiroin: Sort of expected I guess (abstract fs blocks vs concrete block device ones / whatnot).

However, where did you find that "quota uses BLOCK_SIZE?"
I took a closer look at "man quotactl" and it talks about plain "512 byte blocks" - which now makes me wonder about any block size constant that might be 1024 ...

Tomin1: [From Stack overflow](https://stackoverflow.com/questions/2506288/detect-block-size-for-quota-in-linux) I probably should use that glibc macro though.

That reference to 512 in man quotactl seems to be in the section for XFS which has different API so I'm not sure if that applies here.

sailfishusermanager.cpp
Tomin1: Erm, no? Or I'm not sure what you mean but findHomeDevice() returns path to block device and not the mount point. That USER_HOME.arg("") inside the function is maybe a little stupid though.

sailfishusermanager.cpp
spiiroin: That stack overflow answer points to macros in sys/quota.h that use BLOCK_SIZE.
Except that ours uses literal 1024.
Anyway, using BLOCK_SIZE seems legit afaict -> that roundtrip is needed, but maybe one should add a comment about needing to normalize block counts / stte to keep the next reader(s)  informed?

sailfishusermanager.cpp
spiiroin: To clarify: I meant that in this method we effetively arrive to "/home" via two ways: one is literal string, another one uses USER_HOME variable/constant -> works, but making also statfs() use value derived from USER_HOME would ensure they refer to the same thing also in the future.

Tomin1: I see and yes unifying them would be better.

sailfishusermanager.cpp
Tomin1: I wonder if I should also check the value but I guess the code "conveniently" sets the limit to 0 (i.e. not set) if it fails now. Then again it does that silently so maybe not so great?

sailfishusermanager.cpp
Tomin1: Well, I updated it to use the macro. I think it should be good that way and fairly clear that it needs conversion even without a comment. If you think it needs some comment still (like which ones are in bytes and which are in blocks), please tell. Also changed it to use f_frsize instead of f_bsize as that's what statvfs man page says to be the unit. Calculations are now in file system blocks instead of bytes but MAXIMUM_QUOTA_LIMIT is still bytes as it needs to be adjusted by f_frsize.

sailfishusermanager.cpp
spiiroin: In this case: the sizes and counts have probably always been there -> does not really matter -> feel free to leave it as-is.

Also normally kernel and libc can be assumed to be in sync about buffer sizes and content.

But to avoid spending time e.g. wondering why zypper dup in sbox ends up having every 37th or so package failing to install because target file system is supposedly read-only [due to stack noise in f_flag] -> I just zero-initialize these things ;-)

spiiroin (approved):

sailfishusermanager.cpp
spiiroin: yes, using macro from relevant header = ought to be good enough hint about what is going on.

PR #21, opened 2020-06-24
[usermanager] Inform UI if user switch fails. Fixes JB#49966

mkenttala:


sailfishusermanager.cpp
```diff
@@ -577,10 +578,11 @@ void SailfishUserManager::onCreatingJobFailed(SystemdManager::JobList &remaining
             // session systemd is stopped but autologin is still up and it wasn't brought down
             // TODO: What to do?
             qCWarning(lcSUM) << "Could not stop autologin, unhandled error case";
+            emit currentUserChangeFailed(m_switchUser);
```
spiiroin, line 650: that "unhandled error case" logging above - are we now handling the error case i.e. should the log message be adjusted?

sailfishusermanager.cpp
```diff
@@ -556,11 +555,13 @@ void SailfishUserManager::onUnitJobFailed(SystemdManager::Job &job, SystemdManag
         // autologind was started but starting user@.service failed, probably because it was already starting
         qCWarning(lcSUM) << "Starting session systemd failed, is it already starting?";
         m_switchUser = 0;
+    } else {
+        // Inform UI
+        emit currentUserChangeFailed(m_switchUser);
```
spiiroin, line 628: Just wondering... do we get notifications only about jobs started from here? Or can some random unrelated job failure lead into this 'else' block and trigger user switch failure actions?

sailfishusermanager.cpp
Tomin1: Emitting that signal doesn't really do anything useful here as user session is mostly stopped at this point. I don't see this signal doing any harm (if not any good either) but I would not call it handled with just this.

sailfishusermanager.cpp
mkenttala: Been testing this a while and at least so far been unable to get errors.

Tomin1: Those four cases above should handle all possible cases. This else block is never reached as only those jobs that are started by user-managerd are handled here.

That said, this failed signal was before there to tell that switching user failed although there was nothing listening to it, it was clearly about failures. Now it's repurposed for this case that's more about canceling user switch than stopping it. I think we could add a boolean to the signal to indicate whether it was actual failure or of user switching was canceled (i.e. nothing happened) if we want to use the same signal.

mkenttala: Dropped the else block. Of course now when testing I cannot get these to fail so this is more or less just guesswork. I think this signal is still about errors, now there's just someone listening too and showing it in UI. In my book it's an error if nothing happens when requesting user switch.

Tomin1: > In my book it's an error if nothing happens when requesting user switch.

I agree. That's why there was the TODO comment that it should be signaled some way in that particular spot. Also in the bug report there is a suggestion of what kind of signal it could be. I just think that in these two cases we should make some difference there so UI can act the way it needs to. I mean the other failures are really fatal but this one is not so much.

sailfishusermanager.cpp
```diff
@@ -537,11 +537,12 @@ void SailfishUserManager::onUnitJobFinished(SystemdManager::Job &job)
...
 
 void SailfishUserManager::onUnitJobFailed(SystemdManager::Job &job, SystemdManager::JobList &remaining) {
-    emit currentUserChangeFailed(m_switchUser);
     if (job.type == SystemdManager::StopJob && job.unit == USER_SERVICE.arg(m_currentUid)) {
         // session systemd is fubar, autologin is probably still up
         qCWarning(lcSUM) << "Unit failed while stopping session, trying to continue";
         m_systemd->addUnitJobs(remaining); // Try to continue anyway
+        // Inform UI
+        emit currentUserChangeFailed(m_switchUser);
     } else if (job.type == SystemdManager::StopJob && job.unit == AUTOLOGIN_SERVICE.arg(m_currentUid)) {
```
Tomin1: Why this block doesn't emit currentUserChangeFailed? If it should, then that signal could be emited like it was emit before.

mkenttala: If just this autologin stop fails but continue works then the change does not actually fail? Didn't manage to actually test this.

Tomin1: Ah okay yes, then the one above shouldn't emit it either.

Tomin1 (approved):

PR #23, opened 2020-07-01
[user-managerd] Add support for guest user. Fixes JB#50231

mkenttala:
So the idea in this one is if setCurrentUser 111111 is called then the guest user is created and switched to. Next setCurrentUser will delete the user. That uid is just something that I came up and open for debate, unlikely that a device will use more than 11K users :)

### DIFF
--- a/rpm/user-managerd.spec
+++ b/rpm/user-managerd.spec
@@ -1,5 +1,5 @@
 Name: user-managerd
-Version: 0.0.1
+Version: 0.5.17
 Release: 1
 Summary: Sailfish user manager daemon
 License: BSD
@@ -28,7 +28,7 @@ Requires: user-managerd
 %files
 %defattr(-,root,root,-)
 %{_bindir}/%{name}
-%{_unitdir}/*.service
+%{_unitdir}/*
 %{_datadir}/dbus-1/system-services/*.service
 %{_sysconfdir}/dbus-1/system.d/*.conf
 %{_sbindir}/userdel_local.sh
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -39,6 +39,9 @@ const char *GROUPS_USER = "USER_GROUPS";
 const auto GROUPS_IDS_FILE = QStringLiteral("/usr/share/sailfish-setup/group.ids");
 const auto SKEL_DIR = QStringLiteral("/etc/skel");
 const auto USER_HOME = QStringLiteral("/home/%1");
+const auto GUEST_USER = QStringLiteral("sailfish-guest");
+const auto USER_HOME_STARTUPWIZARD_DONE = QStringLiteral("%1/.jolla-startupwizard-done");
+const auto USER_HOME_USERSESSION_DONE = QStringLiteral("%1/.jolla-startupwizard-usersession-done");
 const int HOME_MODE = 0700;
 const int QUIT_TIMEOUT = 60 * 1000; // One minute quit timeout
 const int SWITCHING_DELAY = 1000; // One second time before changing currentUser
@@ -47,6 +50,7 @@ const int OWNER_USER_UID = 100000;
 const auto DEFAULT_TARGET = QStringLiteral("default.target");
 const auto USER_SERVICE = QStringLiteral("user@%1.service");
 const auto AUTOLOGIN_SERVICE = QStringLiteral("autologin@%1.service");
+const auto GUEST_HOME_SERVICE = QStringLiteral("tmp-guest_home.mount");
 const auto ENVIRONMENT_FILE = QStringLiteral("/etc/environment");
 const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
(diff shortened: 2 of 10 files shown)


Tomin1:
How does this handle enabled state of the feature? Or should we store that separately somehow?

mkenttala:
Since it was requested to add enable method it's now needed to first call "enableGuestUser true" and then "setCurrentUser 111111".

sailfishusermanager.cpp
```diff
@@ -777,3 +789,65 @@ void SailfishUserManager::removeFromGroups(uint uid, const QStringList &groups)
         }
     }
 }
+
+void SailfishUserManager::enableGuestUser(bool enable)
```
Tomin1, line 831: This is now duplicating code from addUser quite a bit. Would be feasible to combine those common parts?

sailfishusermanager.cpp
```diff
@@ -777,3 +795,35 @@ void SailfishUserManager::removeFromGroups(uint uid, const QStringList &groups)
...
+
+    struct passwd *pwd = getpwuid(SAILFISH_USERMANAGER_GUEST_UID);
+
+    if (enable) {
+        if (pwd) {
+            // Already enabled
+            return;
+        }
+
+        if (!addSailfishUser(GUEST_USER, GUEST_USER, SAILFISH_USERMANAGER_GUEST_UID))
```
Tomin1: I think the "real" name should be an empty string since we want to localize it.
Edit: Technically that doesn't prevent us from localizing but it would be a little simpler elsewhere.

sailfishusermanager.cpp
```diff
@@ -39,6 +39,7 @@ const char *GROUPS_USER = "USER_GROUPS";
 const auto GROUPS_IDS_FILE = QStringLiteral("/usr/share/sailfish-setup/group.ids");
 const auto SKEL_DIR = QStringLiteral("/etc/skel");
 const auto USER_HOME = QStringLiteral("/home/%1");
+const auto GUEST_USER = QStringLiteral("sailfishguest");
```
Tomin1: This has a problem. It can collide with user selected user name (if they create user named Sailfish Guest for example). Can it be named sailfish-guest?

sailfishusermanager.cpp
```diff
@@ -204,7 +205,10 @@ bool SailfishUserManager::makeHome(const QString &user)
         return false;
     }
 
-    QString destination = QString(USER_HOME).arg(user);
+    QString destination;
+    if (pw->pw_uid == SAILFISH_USERMANAGER_GUEST_UID) destination = SAILFISH_USERMANAGER_GUEST_HOME;
```
Tomin1: If (and else) body should be on new line even if it's pretty short.

sailfishusermanager.cpp
```diff
@@ -397,18 +407,18 @@ void SailfishUserManager::removeUser(uint uid)
     m_exitTimer->start();
 
     if (!removeHome(uid)) {
-        sendErrorReply(QStringLiteral(SailfishUserManagerErrorHomeRemoveFailed), QStringLiteral("Removing user home failed"));
-        return;
```
Tomin1, line 401: Why was this return removed? Also this seems like it could be `if (!removeHome(uid) && calledFromDBus()) { /* ... */ }`

mkenttala: Removed because it prevented removing user without existing home, like "/tmp/guest_home" gets deleted in reboot.

Tomin1: Shouldn't it still return if that failure happens and the user removed is not guest user?

Tomin1: Anyway it feels odd if it emits an error but continues possibly succeeding and emitting also the success case.

rpm/user-managerd.spec
```diff
@@ -1,5 +1,5 @@
 Name: user-managerd
-Version: 0.0.1
+Version: 0.5.17+jb50231
```
rainemak: Version: 0.5.17 is enough

monich: I would say 0.5 is enough unless somebody is planning to maintain the spec version in sync with the reality.

mkenttala: Version was updated because for some reason obs feature branch needed it to use proper version. Hopefully normal obs build uses the tag version as usual.

monich: Well, I do maintain those for my packages but it's totally up to the maintainer. If noone it going to sync those with tags then it's clearly better to have an obviously bugus value there, like 0.0.1

rpm/user-managerd.spec
```diff
@@ -28,7 +28,7 @@ Requires: user-managerd
 %files
 %defattr(-,root,root,-)
 %{_bindir}/%{name}
-%{_unitdir}/*.service
+%{_libdir}/systemd/system/*
```
rainemak: This should be new systemd unit dir macro.

rainemak: Now I understand what was the issue that you saw.

> %{_unitdir}/*

Will work but ```%{_unitdir}/*.service``` just having will not as there is also one .mount

rainemak: So, please change to %{_unitdir}/*

user-managerd.pro
```diff
@@ -45,12 +45,13 @@ DISTFILES += \
...
     rpm/user-managerd.spec \
+    tmp-guest_home.mount \
     userdel_local.sh
 
 target.path = /usr/bin/
 
-systemd.files = dbus-$${DBUS_SERVICE_NAME}.service
-systemd.path = /usr/lib/systemd/system/
+systemd.files = dbus-$${DBUS_SERVICE_NAME}.service tmp-guest_home.mount
+systemd.path = /lib/systemd/system/
```
rainemak: $$[QT_INSTALL_LIBS]/systemd/system/

New files should be going all to the /usr/lib/systemd/system

rainemak:
LGTM :+1:

PR #25, opened 2020-07-10
[usermanager] Disable SUW for guest user. Contributes to JB#50233

mkenttala:


sailfishusermanager.cpp
```diff
@@ -196,7 +198,7 @@ bool SailfishUserManager::copyDir(const QString &source, const QString &destinat
     return true;
 }
 
-bool SailfishUserManager::makeHome(const QString &user)
+bool SailfishUserManager::makeHome(const QString &user, bool suwDone)
```
rainemak: Am I missing something as  I don't see makeHome() being called so that suwDone is true.

rainemak (approved):
LGTM :+1:

PR #26, opened 2020-07-21
[user-managerd] Fix for too great guest uid. Fixes JB#50558

mkenttala:
At least Sailfish X crashes lipstick if guest uid 111111.


rainemak (approved):
LGTM

PR #28, opened 2020-08-04
[user-managerd] Sort removal scripts and expect .sh extension. JB#50398

Tomin1:
This is a good habit to do. Also mentioned in #16 comments.


spiiroin:
:+1:

PR #29, opened 2020-08-10
[user-managerd] Don't remove guest on logout. Fixes JB#50758 

Tomin1:
Don't remove guest user on logout so that it stays enabled. Also instead
of manually handling guest user data deletion and home directory
mounting or creation, leave that to systemd units. Add unit that handles
SUW files whenever guest user logs in.

Log all error messages to journal. Previously some were just signaled 
over D-Bus but not logged. Much easier to debug now.

Remove user even if some group removal fails. Otherwise this could
result in some odd situations that user just can't be removed from UI.

### DIFF
--- a/libuserhelper.cpp
+++ b/libuserhelper.cpp
@@ -248,20 +248,24 @@ bool LibUserHelper::removeUser(uint uid)
             error = nullptr;
         }
 
-        if (rv && lu_group_lookup_id(context, lu_ent_get_first_id(ent, LU_GIDNUMBER), entGroup, &error)) {
-            if (lu_group_delete(context, entGroup, &error)) {
-                if (!lu_user_delete(context, ent, &error)) {
-                    qCWarning(lcSUM) << "User delete failed:" << lu_strerror(error);
-                    lu_error_free(&error);
-                    rv = false;
-                }
-            } else {
+        if (lu_group_lookup_id(context, lu_ent_get_first_id(ent, LU_GIDNUMBER), entGroup, &error)) {
+            if (!lu_group_delete(context, entGroup, &error)) {
                 qCWarning(lcSUM) << "Group delete failed:" << lu_strerror(error);
                 lu_error_free(&error);
+                error = nullptr;
                 rv = false;
             }
         } else {
             qCWarning(lcSUM) << "Could not find group";
+            if (error) {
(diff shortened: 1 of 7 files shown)


libuserhelper.cpp
```diff
@@ -13,6 +13,8 @@
 
 #include <libuser/user.h>
 
+#define LOG_FUNC_INFO << "on function" << __FUNCTION__
```
rainemak: This feels like unnecessary. You can already do all kind of tricks with the logging categories with QT_MESSAGE_PATTERN environment variable and function name should be printed already. You'll find default under /var/lib/environment.

What you could do is that you put the updated QT_MESSAGE_PATTERN to the env upon daemon start. Some good tips can be found for instance from wobog blog: https://woboq.com/blog/nice-debug-output-with-qt.html

rpm/user-managerd.spec
```diff
@@ -51,6 +55,11 @@ make %{?_smp_mflags}
 %install
 make -C build INSTALL_ROOT=%{buildroot} install
 mkdir -p %{buildroot}%{_datadir}/user-managerd/remove.d
+mkdir -p %{buildroot}%{_unitdir}/user@105000.service.wants/
+ln -s ../tmp-guest_home.mount %{buildroot}%{_unitdir}/user@105000.service.wants/
+mkdir -p %{buildroot}%{_unitdir}/autologin@105000.service.wants/
+ln -s ../tmp-guest_home.mount %{buildroot}%{_unitdir}/autologin@105000.service.wants/
+ln -s ../guest_disable_suw.service %{buildroot}%{_unitdir}/autologin@105000.service.wants/
```
rainemak, line 62: Failing to see / find guest_disable_suw.service file. Maybe you forgot to git add it?

rainemak (changes requested):
Please address / clarify questions that were raised.

libuserhelper.cpp
Tomin1: Removed that commit. Thanks!

rpm/user-managerd.spec
Tomin1: Yes, that's what happened. I rewrote the commits and forgot to run git add.

rainemak (approved):
LGTM

PR #31, opened 2020-08-13
Move guest's home to /home

Tomin1:
There isn't any real reason to keep guest's home in /tmp since it lives in a separate file system. Mount it under /home that for example tracker works correctly. Note that the home directory is still different from the /home/$USER scheme as it uses an underscore instead of a dash.

Also I'm refactoring the code a bit. Move group creation inside LibUserHelper::addUser and remove any references to guest user. Instead allow setting uid and home directory path explicitly. This keeps the duties of SailfishUserManager and LibUserHelper a bit cleaner.

### DIFF
--- a/sailfishusermanagerinterface.h
+++ b/sailfishusermanagerinterface.h
@@ -16,7 +16,7 @@
 #define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
 #define SAILFISH_USERMANAGER_DBUS_OBJECT_PATH "/"
 #define SAILFISH_USERMANAGER_GUEST_UID 105000
-#define SAILFISH_USERMANAGER_GUEST_HOME "/tmp/guest_home"
+#define SAILFISH_USERMANAGER_GUEST_HOME "/home/sailfish_guest"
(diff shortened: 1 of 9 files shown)


sailfishusermanagerinterface.h
```diff
@@ -16,7 +16,7 @@
 #define SAILFISH_USERMANAGER_DBUS_INTERFACE "org.sailfishos.usermanager"
 #define SAILFISH_USERMANAGER_DBUS_OBJECT_PATH "/"
 #define SAILFISH_USERMANAGER_GUEST_UID 105000
-#define SAILFISH_USERMANAGER_GUEST_HOME "/tmp/guest_home"
+#define SAILFISH_USERMANAGER_GUEST_HOME "/home/sailfish_guest"
```
rainemak, line 19: So both - and _ are removed from the username when creating home directory name for additional user. Hence, this cannot collide. Correct?

Tomin1: Yes, only letters and numbers are allowed/used.

rainemak (approved):
LGTM

PR #32, opened 2020-09-08
[user-managerd] Add generate user uuid on user creation. Fixes JB#51131

balroggg:
* Add generate uuid on user creation
* Get current user uuid
* Suppress warnings from libuser

### DIFF
--- a/libuserhelper.cpp
+++ b/libuserhelper.cpp
@@ -11,6 +11,7 @@
 #include "logging.h"
 
 #include <libuser/user.h>
+#include <QUuid>
 
 LibUserHelper::LibUserHelper()
 {
@@ -161,6 +162,10 @@ bool LibUserHelper::removeUserFromGroup(const QString &user, const QString &grou
 
 uint LibUserHelper::addUser(const QString &user, const QString& name, uint uid, const QString &home)
 {
+    if (name.contains(',') || name.contains(':')) {
+        qCWarning(lcSUM) << "Invalid user name, comma or colon is not allowed";
+        return 0;
+    }
     struct lu_error *error = nullptr;
     struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, NULL, NULL, &error);
     if (!context) {
@@ -176,7 +181,9 @@ uint LibUserHelper::addUser(const QString &user, const QString& name, uint uid, 
     uint rv = 0;
     struct lu_ent *ent_user = lu_ent_new();
     if (lu_user_default(context, user.toUtf8(), false, ent_user)) {
-        QString fixedName(name);
+        QString nameUuid(name);
+        nameUuid += ',';
+        nameUuid += QUuid::createUuid().toString();
         if (uid) {
             // Explicitly selected uid
             lu_ent_set_id(ent_user, LU_UIDNUMBER, uid);
@@ -187,7 +194,7 @@ uint LibUserHelper::addUser(const QString &user, const QString& name, uint uid, 
         }
 
         lu_ent_set_id(ent_user, LU_GIDNUMBER, gid);
-        lu_ent_set_string(ent_user, LU_GECOS, fixedName.remove(',').toUtf8());
+        lu_ent_set_string(ent_user, LU_GECOS, nameUuid.toUtf8());
 
         if (lu_user_add(context, ent_user, &error)) {
             rv = lu_ent_get_first_id(ent_user, LU_UIDNUMBER);
@@ -293,8 +300,12 @@ bool LibUserHelper::removeUser(uint uid)
     return rv;
 }
 
-bool LibUserHelper::modifyUser(uint uid, const QString &newName)
+bool LibUserHelper::modifyUser(uint uid, const QString &newName) const
 {
+    if (newName.contains(',') || newName.contains(':')) {
+        qCWarning(lcSUM) << "Invalid new user name, comma or colon is not allowed";
+        return false;
+    }
     struct lu_error *error = nullptr;
     struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, NULL, NULL, &error);
     if (!context) {
@@ -307,8 +318,19 @@ bool LibUserHelper::modifyUser(uint uid, const QString &newName)
     struct lu_ent *ent = lu_ent_new();
 
     if (lu_user_lookup_id(context, uid, ent, &error)) {
-        QString fixedName(newName);
-        lu_ent_set_string(ent, LU_GECOS, fixedName.remove(',').remove(':').toUtf8());
+        QString nameUuid(newName);
+        // get uuid from gecos field
+        auto list = QString(lu_ent_get_first_string(ent, LU_GECOS)).split(',');
+        if (list.size() > 1 && !list[1].isEmpty()) {
+            // get old user uuid
+            nameUuid += ',';
+            nameUuid += list[1];
+        } else {
+            // create user uuid if not found in user db
+            if (!nameUuid.endsWith(',')) nameUuid += ',';
+            nameUuid += QUuid::createUuid().toString();
+        }
+        lu_ent_set_string(ent, LU_GECOS, nameUuid.toUtf8());
         if (!lu_user_modify(context, ent, &error)) {
             qCWarning(lcSUM) << "User modify failed:" << lu_strerror(error);
             lu_error_free(&error);
@@ -387,4 +409,56 @@ QStringList LibUserHelper::groups(uint uid)
     lu_end(context);
 
     return rv;
+}
+
+QString LibUserHelper::getUserUuid(uint uid) const
+{
+    struct lu_error *error = nullptr;
+    struct lu_context *context = lu_start(NULL, lu_user, NULL, NULL, NULL, NULL, &error);
+    if (!context) {
+        qCWarning(lcSUM) << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return QString();
+    }
+
+    struct lu_ent *ent = lu_ent_new();
+
+    QString userUuid;
+    if (lu_user_lookup_id(context, uid, ent, &error)) {
+        QString gecos = lu_ent_get_first_string(ent, LU_GECOS);
+        const auto list = gecos.split(',');
+        if (list.size() > 1 && !list[1].isEmpty()) {
+            userUuid = list[1];
+        } else {
+            // try to generate user uuid
+            if (modifyUser(uid, list.count() >= 1 ? list[0] : QString())) {
+                struct lu_ent *ent_new = lu_ent_new();
+                if (lu_user_lookup_id(context, uid, ent_new, &error)) {
+                    gecos = lu_ent_get_first_string(ent_new, LU_GECOS);
+                    lu_ent_free(ent_new);
+                    const auto list = gecos.split(',');
+                    if (list.size() > 1 && !list[1].isEmpty()) {
+                        userUuid = list[1];
+                    }
+                } else {
+                    qCWarning(lcSUM) << "Could not find user after adding uuid";
+                    if (error) {
+                        qCWarning(lcSUM) << "Could not find user after adding uuid" << lu_strerror(error);
+                        lu_error_free(&error);
+                    }
+                }
+            }
+        }
+    } else {
+        qCWarning(lcSUM) << "Could not find user";
+        if (error) {
+            qCWarning(lcSUM) << "Could not find user" << lu_strerror(error);
+            lu_error_free(&error);
+        }
+    }
+
+    lu_ent_free(ent);
+    lu_end(context);
+
+    return userUuid;
 }
--- a/sailfishusermanager.cpp
+++ b/sailfishusermanager.cpp
@@ -695,6 +695,28 @@ uint SailfishUserManager::currentUser()
     return uid;
 }
 
+QString SailfishUserManager::currentUserUuid()
+{
+    const auto uid = currentUser();
+    if (uid == SAILFISH_UNDEFINED_UID) {
+        return QString();
+    }
+    return userUuid(uid);
+}
+
+QString SailfishUserManager::userUuid(uint uid)
+{
+    m_exitTimer->start();
+    const auto userUuid = m_lu->getUserUuid(uid);
+    if (userUuid.isEmpty()) {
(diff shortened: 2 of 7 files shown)


sailfishusermanager.cpp
```diff
@@ -37,6 +37,7 @@ namespace {
 const char *USER_GROUP = "users";
 const char *GROUPS_USER = "USER_GROUPS";
 const auto GROUPS_IDS_FILE = QStringLiteral("/usr/share/sailfish-setup/group.ids");
+const auto USER_UUID_FILE = QStringLiteral("/etc/user_uuid");
```
mkenttala: Is there some reason why passwd gecos field is not used for the uuid? That would be more standardized way of storing user data.

sailfishusermanager.cpp
```diff
@@ -382,6 +391,44 @@ void SailfishUserManager::setUserLimits(uint uid)
...
+    new_content += QUuid::createUuid().toString();
+    new_content += '\n';
+
+    QFile new_file("/etc/user_uuid");
+    if (!new_file.open(QIODevice::WriteOnly | QIODevice::Text))
+        return false;
+
+    QTextStream out(&new_file);
+    out << new_content;
+
```
mkenttala: Missing new_file.close()

sailfishusermanager.cpp
```diff
@@ -695,6 +742,42 @@ uint SailfishUserManager::currentUser()
...
+        uint id  = list[0].toUInt();
+        if ( id != uid ) {
+            continue;
+        }
+        uuid = list[1];
+        break;
+    }
+
+    if (uuid.isEmpty()) {
+        auto message = QStringLiteral("Failed to get current user uuid, cann't find uuid in db");
```
mkenttala: Typo cann't

sailfishusermanager.cpp
balroggg: May be we can do it, for now we store full user name in gecos, we can add custom format for this field. Let me check this.

sailfishusermanager.cpp
balroggg: I think this happens in dtor of qfile object, but if you want I can add this.

sailfishusermanager.cpp
balroggg: @mkenttala I think we can do it. We add some restrictions on full user name(no comma), is it ok?

mkenttala: Yes, sounds good.

balroggg:
Create user uuid in gecos field. Add some checks(, and : symbols) to add/modify user.

sailfishusermanager.cpp
balroggg: This moves to libuser lib.

libuserhelper.cpp
```diff
@@ -388,3 +411,32 @@ QStringList LibUserHelper::groups(uint uid)
...
+        qCWarning(lcSUM) << "Error creating context:" << lu_strerror(error);
+        lu_error_free(&error);
+        return {};
+    }
+
+    struct lu_ent *ent = lu_ent_new();
+
+    QString userUuid;
+    if (lu_user_lookup_id(context, uid, ent, &error)) {
+        userUuid = lu_ent_get_first_string(ent, LU_GECOS);
```
mkenttala: Doesn't this return the whole gecos field like "name,uuid"? Name needs to stripped.

balroggg: Yes, need to strip.

mkenttala (approved):
LGTM

mkenttala:
Should work as advertised. Does @Tomin1 have any objections since you have worked more with multiuser lately?

libuserhelper.cpp
```diff
@@ -161,6 +162,10 @@ bool LibUserHelper::removeUserFromGroup(const QString &user, const QString &grou
 
 uint LibUserHelper::addUser(const QString &user, const QString& name, uint uid, const QString &home)
 {
+    if (name.contains(',') || name.contains(':')) {
+        qCWarning(lcSUM) << "Invalid user name, comma or colon is not allowed";
+        return false;
```
Tomin1: This method's return type is uint but false is boolean. Doesn't seem right to me.

sailfishusermanager.h
```diff
@@ -72,6 +73,7 @@ private slots:
 
 private:
     bool checkAccessRights(uint uid_to_modify);
+    bool setUserUUID(uint uid);
```
Tomin1: I don't see definition for this and I don't exactly see the reason to have this, so maybe this line is a leftover and it should be removed?

libuserhelper.cpp
```diff
@@ -176,7 +181,9 @@ uint LibUserHelper::addUser(const QString &user, const QString& name, uint uid,
     uint rv = 0;
     struct lu_ent *ent_user = lu_ent_new();
     if (lu_user_default(context, user.toUtf8(), false, ent_user)) {
-        QString fixedName(name);
+        QString nameUuid(name);
+        nameUuid += ',';
+        nameUuid += QUuid::createUuid().toString();
```
Tomin1, line 186: This is not necessarily an error but to me it feels like this uuid creation should be inside SailfishUserManager class and the value should be just passed to LibUserHelper that would then just save it.

BTW, at the moment user can change their gecos information with chfn command from util-linux and this allows user to change this uuid as well (now it's saved as office information, option -o).

Tomin1: Just thinking to avoid this problem we should consider either:
- removing chfn
- altering chfn so that we can store extra fields to the end of gecos that can't be changed with chfn (I'm not sure if that can be done securely though)

libuserhelper.cpp
```diff
@@ -307,8 +318,20 @@ bool LibUserHelper::modifyUser(uint uid, const QString &newName)
...
-        lu_ent_set_string(ent, LU_GECOS, fixedName.remove(',').remove(':').toUtf8());
+        QString realName(newName);
+        // get uuid from gecos field
+        auto list = QString(lu_ent_get_first_string(ent, LU_GECOS)).split(',');
+        if (list.size() > 1) {
+            // get old user uuid
+            realName += ',';
+            realName += list[1];
+        }
+        else {
```
Tomin1: Should be on the same line with }.

sailfishusermanager.cpp
```diff
@@ -695,6 +695,19 @@ uint SailfishUserManager::currentUser()
...
 
+QString SailfishUserManager::currentUserUUID()
+{
+    const uint uid = currentUser();
+    const QString uuid = m_lu->getUserUUID(uid);
+    if (uuid.isEmpty()) {
+        auto message = QStringLiteral("Failed to get current user uuid");
+        qCWarning(lcSUM) << message;
+        sendErrorReply(QStringLiteral(SailfishUserManagerErrorGetUuidFailed), message);
+        return {};
```
Tomin1: I would use return QString(); instead. This same thing is elsewhere too.

libuserhelper.cpp
```diff
@@ -307,8 +318,20 @@ bool LibUserHelper::modifyUser(uint uid, const QString &newName)
     struct lu_ent *ent = lu_ent_new();
 
     if (lu_user_lookup_id(context, uid, ent, &error)) {
-        QString fixedName(newName);
-        lu_ent_set_string(ent, LU_GECOS, fixedName.remove(',').remove(':').toUtf8());
+        QString realName(newName);
+        // get uuid from gecos field
+        auto list = QString(lu_ent_get_first_string(ent, LU_GECOS)).split(',');
+        if (list.size() > 1) {
```
Tomin1: if (list.size() > 1 && !list[1].isEmpty()) {

libuserhelper.cpp
Tomin1: Ehm. Sorry, it's actually prevented already on Sailfish. You may ignore this.
`chfn: /etc/login.defs: CHFN_RESTRICT does not allow any changes`

libuserhelper.cpp
balroggg: Yeah, need to fix.

libuserhelper.cpp
balroggg: Ok, good to know.

sailfishusermanager.h
balroggg: Yes, maybe some old code. I remove this.

sailfishusermanager.cpp
balroggg: Ok

libuserhelper.cpp
balroggg: Ok

libuserhelper.cpp
balroggg: Ok

Tomin1 (approved):
LGTM. Just an idea but when it fetches current uuid should it also generate a new one if it doesn't exist? That would provide an upgrade path for existing users.

balroggg:
Yes, we can do it, but is it never happens? Because we upgrade from one user to multiuser, and new users already will be created with uuid?

I have some questions from mdm developers, about api to get users uuid. Do we need add something similar like now(get current user uuid) or it's redundant for user-managerd?

Tomin1:
> Yes, we can do it, but is it never happens? Because we upgrade from one user to multiuser, and new users already will be created with uuid?

I was thinking about nemo (now defaultuser, aka Device owner / Admin) that already exists on device. If they don't need this, then it's fine.

> I have some questions from mdm developers, about api to get users uuid. Do we need add something similar like now(get current user uuid) or it's redundant for user-managerd?

I think all access from MDM should go through user-managerd and use its APIs. Does that sound okay? If you need to access other UUIDs than current user's then there should be a D-Bus method for that too.

balroggg:
Ok, I ask mdm devs one more time to be clear.

balroggg:
Yeah, I think we need provide this functionality for consistency(uuid for all users and dbus api for get uuids). I upload changes.

balroggg:
Add generate uuid if not found.
Add get user uuid api.

libuserhelper.cpp
```diff
@@ -388,3 +410,55 @@ QStringList LibUserHelper::groups(uint uid)
...
+
+    QString userUuid;
+    if (lu_user_lookup_id(context, uid, ent, &error)) {
+        QString gecos = lu_ent_get_first_string(ent, LU_GECOS);
+        const auto list = gecos.split(',');
+        if (list.size() > 1 && !list[1].isEmpty()) {
+            userUuid = list[1];
+        } else {
+            // try to generate user uuid
+            if (modifyUser(uid, gecos)) {
```
Tomin1: I think the second argument can't be `gecos` as it may contain `,` characters. It could be `list.count() >= 1 ? list[0] : QString()` I think.

sailfishusermanager.cpp
```diff
@@ -695,6 +695,31 @@ uint SailfishUserManager::currentUser()
     return uid;
 }
 
+QString SailfishUserManager::currentUserUUID()
+{
+    const uint uid = currentUser();
+    const QString uuid = m_lu->getUserUUID(uid);
```
okodron: I guess you can just call `return userUUID(currentUser())` here for code reusage.

okodron (changes requested):

libuserhelper.cpp
balroggg: Yeah, hard to see with things without ut. I fix this.

sailfishusermanager.cpp
balroggg: Yes, we can.

sailfishusermanagerinterface.h
```diff
@@ -30,9 +30,12 @@
 #define SailfishUserManagerErrorUserModifyFailed "org.sailfishos.usermanager.Error.UserModifyFailed"
 #define SailfishUserManagerErrorUserRemoveFailed "org.sailfishos.usermanager.Error.UserRemoveFailed"
 #define SailfishUserManagerErrorGetUidFailed "org.sailfishos.usermanager.Error.GetUidFailed"
+#define SailfishUserManagerErrorGetUuidFailed "org.sailfishos.usermanager.Error.GetUuidFailed"
```
Tomin1, line 33: For consistency, please name this similarly to others, i.e. UUID instead of Uuid.

libuserhelper.cpp
```diff
@@ -307,8 +318,19 @@ bool LibUserHelper::modifyUser(uint uid, const QString &newName)
     struct lu_ent *ent = lu_ent_new();
 
     if (lu_user_lookup_id(context, uid, ent, &error)) {
-        QString fixedName(newName);
-        lu_ent_set_string(ent, LU_GECOS, fixedName.remove(',').remove(':').toUtf8());
+        QString realName(newName);
```
Tomin1: In the other place similar variable is called nameUuid, I think it would be good to use the same name in both places. It's not totally necessary though but would be easier to see the connection in future.

Tomin1 (commented):
I think after this the code is good so could you squash the commits so that there is just one commit to merge. Thanks!

sailfishusermanagerinterface.h
balroggg: Ok, or may be Uuid like qt class QUuid?

libuserhelper.cpp
balroggg: Ok np

sailfishusermanagerinterface.h
Tomin1: That'd be okay too but then change all instances of UUID to Uuid here and also on D-Bus interfaces.

balroggg: Ok done

sailfishusermanagerinterface.h
```diff
@@ -30,10 +30,12 @@
...
 #define SailfishUserManagerErrorUserNotFound "org.sailfishos.usermanager.Error.UserNotFound"
 #define SailfishUserManagerErrorAddToGroupFailed "org.sailfishos.usermanager.Error.AddToGroupFailed"
 #define SailfishUserManagerErrorRemoveFromGroupFailed "org.sailfishos.usermanager.Error.RemoveFromGroupFailed"
 
+
```
Tomin1: One extra line.

Tomin1 (approved):
One nitpick but LGTM.

sailfishusermanagerinterface.h
balroggg: Ok, fixed

PR #33, opened 2020-09-21
[user-managerd] Add documentation. Contributes to JB#51230

Tomin1:
Add documentation for org.sailfishos.usermanager API and sailfishusermanagerinterface.h header. Moved some of the content in README.md to index.qdoc. I chose to mark those interfaces to internal that are not relevant for D-Bus API and are used only by main.

Also reorganized to subdirectories. Moved sources to src/ and systemd or D-Bus related things to service/ to
keep this clean and tidy. Added doc/ for documentation. Reworked .pc file generation and added .gitignore.


Tomin1:
Building the package generates documentation to doc/html/ so you can check it there. Also if anyone knows how to exclude this "`Header:	#include <SailfishUserManager>`" text from SailfishUserManager class' page, I'd like to add that here. That header name is wrong and also non-public (i.e. not included in devel subpackage).

doc/index.qdoc
```diff
@@ -0,0 +1,50 @@
...
+  Daemon for handling Sailfish OS device users.
+
+  This daemon provides D-Bus API for adding, removing and editing additional
+  Sailfish OS device users. The main system user is created during the first
+  boot of the device. This user can then use this API to add additional users
+  which have access only to some of the available functionality and services.
+
+  \section1 Using Sailfish User Manager Daemon
+
+  Sailfish User Manager Daemon is used through \c org.sailfishos.usermanger
```
mkenttala: typo usermanger :)

mkenttala (approved):

doc/index.qdoc
Tomin1: Thanks, fixed. :smile:

src/src.pro
```diff
@@ -0,0 +1,68 @@
...
+
+target.path = /usr/bin/
+
+pkgconfig.files = sailfishusermanager.pc
+pkgconfig.path = $$[QT_INSTALL_LIBS]/pkgconfig
+pkgconfig.CONFIG = no_check_exist
+pkgconfig.commands = $$QMAKE_STREAM_EDITOR \
+    -e 's~@VERSION@~$$VERSION~' \
+    -e 's~@LIBDIR@~$$[QT_INSTALL_LIBS]~' \
+    -e 's~@INCLUDEDIR@~$$include.path~' \
```
Tomin1: Oops, this line doesn't actually work.

PR #34, opened 2020-09-25
[user-managerd] Update to use group_ids.env. JB#51329

Tomin1:
File format for specifying groups for users changed in sailfish-setup,
this changes user-managerd to support that. Also add users to the rest
of the groups even in case of an error and also slightly better error
reporting in that case.

Also change privileged check to check also that if user belongs to
privileged group, they should be allowed. This is for sailfish-mdm user.

Requires: https://github.com/sailfishos/sailfish-setup/pull/20


Tomin1:
Renamed group.ids in https://github.com/sailfishos/sailfish-setup/pull/20, updated this accordingly.

rainemak (approved):
LGTM

PR #35, opened 2021-01-14
[user-managerd] Set usb-mode on user change. Contributes to JB#50877

salmelamike:
Set usb-mode to ask if MTP mode is enabled while changing user.


Tomin1:
@salmelamike If this is not needed anymore, please close.

PR #36, opened 2021-02-02
Fixes to group management

Tomin1:
Check that groups added with addToGroups or removed with
removeFromGroups are prefixed with sailfish- or account- prefix.

Change removeGroup to take gid instead of group name and move group
removal on failure inside LibUserHelper::addUser() as it knows better
how to deal with it.

Check in SailfishUserManager::addUser() that the group name is not
reserved. Also treat already existing home directory the same way
(increment the number at the end of username).

### DIFF
--- a/src/libuserhelper.cpp
+++ b/src/libuserhelper.cpp
@@ -54,7 +54,7 @@ uint LibUserHelper::addGroup(const QString &group, int gid)
     return rv;
 }
 
-bool LibUserHelper::removeGroup(const QString &group)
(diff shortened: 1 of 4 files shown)


mkenttala (approved):
LGTM

izh1979:
LGTM.

Tomin1:
Fixed a mistake with the first check I added. Also added that check to removeFromGroups for symmetry. Now I tested that this is actually working for all changes.

mkenttala (approved):

PR #37, opened 2021-03-24
[usermanager] Add support for user creation scripts. Contributes to JB#53668

MrBam:
Add support for user creation scripts with root rights when new user is creating.
Newuser oneshot are working only with new user rights, that not enough to create system folders.


spiiroin:
The bug refs in commit titles need to be: "... JB#53668" (instead of "jb53668")

src/sailfishusermanager.cpp
```diff
@@ -361,7 +362,18 @@ uint SailfishUserManager::addSailfishUser(const QString &user, const QString &na
...
+    creation.setSorting(QDir::NoSort);
+    creation.setFilter(QDir::Files | QDir::Executable);
+
+    auto entryList = creation.entryList();
+
+    QCollator collator;
+    collator.setNumericMode(true);
+    collator.setCaseSensitivity( Qt::CaseInsensitive );
+
+    std::sort(entryList.begin(), entryList.end(), collator);
```
spiiroin: This fine-tuned sorting makes adding and removing users behave differently. And this filter+sort variant seems to have problems that remove avoids e.g. using "*.sh" pattern ensures random files that were accidentally installed are not executed (say, editor backup files, either due to on-device ad-hoc testing or hiccups in rpm packaging).

I'd suggest:

- instead of inlining execute loop here, implement and use createUserFiles()  analogous to already existing removeUserFiles()
- would not hurt if both would use common code for the execute-scripts-in-dir bit
- but in any case: if more fine tuned sorting is needed, it should be used both for create and remove scripts

spiiroin (changes requested):

src/sailfishusermanager.cpp
```diff
@@ -358,6 +360,25 @@ uint SailfishUserManager::addSailfishUser(const QString &user, const QString &na
         return 0;
     }
 
+    // Execute user creation scripts
+    QDir creation(USER_CREATE_SCRIPT_DIR);
+    creation.setSorting(QDir::NoSort);
+    creation.setFilter(QDir::Files | QDir::Executable);
+
+    auto entryList = creation.entryList();
```
Tomin1: You could have just `creation.entryList(QStringList() << "*.sh", QDir::Files | QDir::Executable, QDir::Name)`. That sorts by name. Just like in `removeUserFiles()`.

MrBam: QDir::Name will sort like this 
1.sh
10.sh
2.sh

instead 
1.sh
2.sh
10.sh

Tomin1: Yes. Then again it's pretty much a convention to name such files like `01-foo.sh`, `02-bar.sh` and `10-foobar.sh` which maintains the expected order even with `QDir::Name`. Currently none of the files we have is named with a number at the beginning though.

What actually matters here is that the order is deterministic, i.e. it doesn't change from device to device so the behaviour is always the same. Without any sorting the order depeds on for example the order the packages where installed and if there is some dependency between the files, then any bugs that occure might not be easily reproducible. It's less important to ensure that the order is natural than to ensure that it's always the same.

PR #38, opened 2021-03-25
[usermanager] Add support for user creation scripts and move remove scripts to functions. Contributes to JB#53668

MrBam:
Add support for user creation scripts with root rights when new user is creating.
Newuser oneshot are working only with new user rights, that not enough to create system folders.

adding sorting of file for remove scripts

### DIFF
--- a/src/sailfishusermanager.cpp
+++ b/src/sailfishusermanager.cpp
@@ -18,6 +18,7 @@
 #include <QFile>
 #include <QDir>
 #include <QString>
+#include <QCollator>
 
 #include <errno.h>
 #include <grp.h>
@@ -55,6 +56,7 @@ const QByteArray LAST_LOGIN_UID_KEY("LAST_LOGIN_UID=");
 const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
 const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/remove.d");
+const auto USER_CREATE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/create.d");
 const quint64 MAXIMUM_QUOTA_LIMIT = 2000000000ULL;
 const auto SAILFISH_GROUP_PREFIX = QStringLiteral("sailfish-");
 const auto ACCOUNT_GROUP_PREFIX = QStringLiteral("account-");
@@ -332,6 +334,24 @@ uint SailfishUserManager::addUser(const QString &name)
     return addSailfishUser(user, name);
 }
 
+void SailfishUserManager::executeScripts(uint uid, const QString &directory)
+{
+    QDir creation(directory, "*.sh", QDir::NoSort, QDir::Files | QDir::Executable);
+
+    auto entryList = creation.entryList();
+
+    QCollator collator(QLocale::C);
+    collator.setNumericMode(true);
+
+    std::sort(entryList.begin(), entryList.end(), collator);
+
+    for (const QString &entry : entryList) {
+        int exitCode = QProcess::execute(directory + '/' + entry, QStringList() << QString::number(uid));
+        if (exitCode)
+            qCWarning(lcSUM) << "User scripts" << directory + '/' + entry << "returned:" << exitCode;
+    }
+}
+
 uint SailfishUserManager::addSailfishUser(const QString &user, const QString &name, uint userId, const QString &home)
 {
     uint uid = m_lu->addUser(user, name, userId, home);
@@ -358,6 +378,9 @@ uint SailfishUserManager::addSailfishUser(const QString &user, const QString &na
         return 0;
     }
 
+    // Execute user creation scripts
+    executeScripts(uid, USER_CREATE_SCRIPT_DIR);
+
     setUserLimits(uid);
 
     SailfishUserManagerEntry entry;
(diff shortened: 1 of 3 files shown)


src/sailfishusermanager.cpp
```diff
@@ -332,6 +334,26 @@ uint SailfishUserManager::addUser(const QString &name)
...
 }
 
+int SailfishUserManager::creationScriptsUserFiles(uint uid)
+{
+    // Execute user creation scripts
+    QDir creation(USER_CREATE_SCRIPT_DIR, "*.sh", QDir::NoSort, QDir::Files | QDir::Executable);
+
+    auto entryList = creation.entryList();
+
+    QCollator collator;
```
Tomin1: I think locale aware comparison is not something we want to do here. It would be best if this compared the strings the same way on every device. From [documentation](https://doc.qt.io/archives/qt-5.6/qcollator.html#QCollator):

> Constructs a QCollator from locale. If locale is not specified the system's default locale is used.

Of course if you want to use QCollator, this should probably define C as locale.

src/sailfishusermanager.cpp
```diff
@@ -332,6 +334,26 @@ uint SailfishUserManager::addUser(const QString &name)
...
+    collator.setNumericMode(true);
+    collator.setCaseSensitivity( Qt::CaseInsensitive );
```
Tomin1: I think this probably introduces some non-determinism here. So if two files only differ by case, they can be returned by `entryList()` in either order and if this is stable sort, then it can vary between devices. Unlikely case but possible and undesired. Also if the sort is not stable, then it doesn't really matter since it's undesired anyway. I would say we want this to be case sensitive so that all names are different from each other.

src/sailfishusermanager.cpp
```diff
@@ -379,9 +403,17 @@ int SailfishUserManager::removeUserFiles(uint uid)
         qCWarning(lcSUM) << "Removing user environment directory failed";
 
     // Execute user removal scripts
-    QDir removal(USER_REMOVE_SCRIPT_DIR);
-    for (const QString &entry : removal.entryList(QStringList() << "*.sh",
-                                                  QDir::Files | QDir::Executable, QDir::Name)) {
+    QDir removal(USER_CREATE_SCRIPT_DIR, "*.sh", QDir::NoSort, QDir::Files | QDir::Executable);
```
Tomin1: This should be `USER_REMOVE_SCRIPT_DIR`. Also I think what @spiiroin meant with his previous comment in the other PR was that these both code paths could be combined since they need to only differ by the directory used here. So something like `void executeScripts(uint uid, const QString &directory);`

Tomin1 (changes requested):

src/sailfishusermanager.cpp
MrBam: fix it

src/sailfishusermanager.cpp
```diff
@@ -332,6 +334,25 @@ uint SailfishUserManager::addUser(const QString &name)
...
 
+void SailfishUserManager::executeScripts(uint uid, const QString &directory)
+{
+    QDir creation(directory, "*.sh", QDir::NoSort, QDir::Files | QDir::Executable);
+
+    auto entryList = creation.entryList();
+
+    QCollator collator;
+    collator.setNumericMode(true);
+    collator.setLocale( QLocale::C );
```
Tomin1: To avoid intializing twice, you could set locale already in constructor.
`QCollator collator(QLocale::C);`

Tomin1 (commented):
Just that one little thing

Tomin1 (approved):

src/sailfishusermanager.cpp
```diff
@@ -332,6 +334,24 @@ uint SailfishUserManager::addUser(const QString &name)
     return addSailfishUser(user, name);
 }
 
+void SailfishUserManager::executeScripts(uint uid, const QString &directory)
+{
+    QDir creation(directory, "*.sh", QDir::NoSort, QDir::Files | QDir::Executable);
```
spiiroin, line 339: nit: Now that this is used for both create and remove "creation" is a bit of a misnomer.

spiiroin (changes requested):
Logic wise looks ok to me now.

But having three commits for this... if it is not too much trouble, could you please squeeze the commits.

src/sailfishusermanager.cpp
MrBam: and how it's should be better written ?

MrBam:
> Logic wise looks ok to me now.
> 
> But having three commits for this... if it is not too much trouble, could you please squeeze the commits.

done it

src/sailfishusermanager.cpp
spiiroin: Well, I do not have an obvious candidate, just a bit irked to think removal operation would iterate creation list... It is a QDir by type but is used for locating script files -> for "create" purpose it could have been "creationScripts", and for removal "removalScripts" might have been appropriate -> now that it is generic enough to deal with both of those / something else too -> droping the parts that hint to specific purpose would leave: "scripts"?

MrBam: ok. I change to executeScripts. Any change needed ?

spiiroin: I did not cache the previous PR version, so I can't figure out what changed.  Did you change the method name? I meant the variable "creation". In any case it was cosmetic to begin with, so feel free to leave it as it is now.

spiiroin (approved):

MrBam:
@Tomin1 @spiiroin any questions ? can it be merged ?

PR #39, opened 2021-04-21
Include linux/quota.h

PureTryOut:
Otherwise if_dqblk is considered an incomplete type


pvuorela:
Looks like the same musl shortcoming as discussed in https://git.sailfishos.org/mer-core/nemo-qml-plugin-systemsettings/merge_requests/174

PR #40, opened 2021-09-06
[user-managerd] Add script execution on user pre-switch. JB#55415

rinatdobr:
Added ~~new two points of the~~ script execution:
* `user pre switch` is called before switching to the another user.
* ~~`user session` on the new user session: user boot up or after switch to the another user. That point was added via new command line arg `--session` because user-managerd does not catch that state.~~

### DIFF
--- a/rpm/user-managerd.spec
+++ b/rpm/user-managerd.spec
@@ -45,6 +45,7 @@ Summary: Sailfish User Manager Daemon documentation
 %{_sbindir}/userdel_local.sh
 %{_datadir}/user-managerd/remove.d
 %{_datadir}/user-managerd/create.d
+%{_datadir}/user-managerd/pre-switch.d
 
 %files devel
 %{_prefix}/include/sailfishusermanager
@@ -65,6 +66,7 @@ make %{?_smp_mflags}
 
 mkdir -p %{buildroot}%{_datadir}/user-managerd/remove.d
 mkdir -p %{buildroot}%{_datadir}/user-managerd/create.d
+mkdir -p %{buildroot}%{_datadir}/user-managerd/pre-switch.d
 mkdir -p %{buildroot}%{_unitdir}/user@105000.service.wants/
 ln -s ../home-sailfish_guest.mount %{buildroot}%{_unitdir}/user@105000.service.wants/
 mkdir -p %{buildroot}%{_unitdir}/autologin@105000.service.wants/
--- a/src/sailfishusermanager.cpp
+++ b/src/sailfishusermanager.cpp
@@ -57,6 +57,7 @@ const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
 const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/remove.d");
 const auto USER_CREATE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/create.d");
+const auto USER_PRE_SWITCH_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/pre-switch.d");
 const quint64 MAXIMUM_QUOTA_LIMIT = 2000000000ULL;
 const auto SAILFISH_GROUP_PREFIX = QStringLiteral("sailfish-");
 const auto ACCOUNT_GROUP_PREFIX = QStringLiteral("account-");
@@ -701,6 +702,9 @@ void SailfishUserManager::setCurrentUser(uint uid)
         if (!m_systemd) {
             initSystemdManager();
         }
+
+        executeScripts(m_currentUid, USER_PRE_SWITCH_SCRIPT_DIR);
+
         qCDebug(lcSUM) << "Switching user from" << m_currentUid << "to" << m_switchUser << "now";
         m_systemd->addUnitJobs(SystemdManager::JobList()
                                << SystemdManager::Job::stop(USER_SERVICE.arg(m_currentUid))


src/sailfishusermanager.cpp
```diff
@@ -465,6 +467,16 @@ int SailfishUserManager::removeUserFiles(const char *user)
     return EXIT_FAILURE;
 }
 
+/*!
+  \brief Process user session.
+ */
+int SailfishUserManager::processSession(const char *uid)
+{
+    executeScripts(QString(uid).toUInt(), USER_SESSION_SCRIPT_DIR);
```
Tomin1: I think you should get the same effect if you put this call inside`onUnitJobFinished()` so that there is no need to add extra systemd units or running another copy of user-managerd with special arguments.

I suggest that you either have this call after `autologin@<uid>.service` job finishes (i.e. `autologin@.service` has started) or after `user@<uid>.service` start job finishes so it's actually a "late" service. The former would need a new branch to be added of course, e.g. `job.type == SystemdManager::StartJob && job.unit == AUTOLOGIN_SERVICE.arg(m_switchUser)`. The latter should work if you just have it after that `currentUserChanged(m_switchUser)` signal which is there already (also consider that "backup plan" in that case).

rpm/user-managerd.spec
```diff
@@ -65,6 +69,8 @@ make %{?_smp_mflags}
 
 mkdir -p %{buildroot}%{_datadir}/user-managerd/remove.d
 mkdir -p %{buildroot}%{_datadir}/user-managerd/create.d
+mkdir -p %{buildroot}%{_datadir}/user-managerd/session.d
```
Tomin1: I'm also wondering if these names should have some symmetry, like `pre-switch.d` and `after-switch.d`. Does that make sense to you?

src/sailfishusermanager.cpp
rinatdobr: `onUnitJobFinished` is fired only when there was an user switch operation. But here is also need to catch user session after device boot.

The service isn't running continuously, it has the exit timer, so I don't think that a `insideonUnitJobFinished` is a good idea to deal with it.

It has already used `user@<uid>.service` to launch the user-managerd to execute scripts.

rpm/user-managerd.spec
rinatdobr: It does not relate to the user switch operation. User-managerd does not detect user change on boot up now.

src/sailfishusermanager.cpp
Tomin1: I see. That is indeed a bit different. How about not involving `user-managerd` at all? You could have a service `user-session-scripts@.service` which would contain something like this:
```ini
[Unit]
Description=Scripts on user session
Before=user@%i.service
PartOf=user@%i.service

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/libexec/user-session-scripts start %i
ExecStop=/usr/libexec/user-session-scripts stop %i
```

And then have that `/usr/libexec/user-session-scripts` contain a little script that would execute the scripts in those directories. Last you would have that systemd conf file so that this is executed when `autologin@<uid>.service` or `user@<uid>.service` starts.

Just saying this, but I think the timer does not affect this. The service is involved in all user switches and it will not stop until they have completed (see `exitTimeout()` for how this is handled). Besides this, executing scripts would block main thread and the timer can not fire while executing them.

rinatdobr: I understand, but wouldn't be it better to bring together all user stuff to the one component and do not spread them among different services/scripts?

Tomin1: We intended user-managerd to be a component that would take part in modifying users and in user switches because those could not be done from UI directly due to insufficient permissions but it wasn't intended to start on bootup. This change would effectively make user-managerd to start on bootup even if there are no scripts to run. I suppose something like ConditionDirectoryNotEmpty setting could prevent it in that case. Anyway since user-managerd is built on Qt it's not very lightweight service to start, which is something to consider.

We have some similar session start and end related scripts as systemd services already. They are in a different repository. The current scripting support in this repository is strictly for user creation or removal which is very much about user management, the primary use for this service, so it's in some ways different from this although maybe not too far.

src/main.cpp
```diff
@@ -16,6 +16,9 @@ int main(int argc, char *argv[])
     QCoreApplication app(argc, argv);
     if (argc > 2 && !strcmp(argv[1], "--removeUserFiles"))
         return SailfishUserManager::removeUserFiles(argv[2]);
+    if (argc == 3 && !strcmp(argv[1], "--session")) {
```
Tomin1: If we choose to support these scripts here, I would prefer this to be something more like "--runSessionStartScripts".

rpm/user-managerd.spec
Tomin1: Ok. Maybe it should be `session-start.d` instead. I'm just thinking what this "session" here means. Existing create and remove are clearly actions (verbs) and relate to users (a kind of default for user-managerd), so somehow having word session to differentiate this from user scripts and then have the action there as well would make sense to me. Anyway, after the directory is in use, it's a lot more work to rename it so it'd be good to at least have a reasonable name there.

Similarly that other one could also have the word session so it's perhaps a bit more clear, so `pre-session-switch.d` maybe.

service/org.sailfishos.usermanager-late@.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Run usermanagerd to execute session scripts
```
Tomin1: Perhaps this should be just "Execute session start up scripts".

service/service.pro
```diff
@@ -6,18 +6,25 @@ DISTFILES += \
     dbus-org.sailfishos.usermanager.service \
     org.sailfishos.usermanager.conf \
     org.sailfishos.usermanager.service \
-    home-sailfish_guest.mount
+    home-sailfish_guest.mount \
+    org.sailfishos.usermanager-late.conf \
+    org.sailfishos.usermanager-late@.service
```
Tomin1: Could we name this differently? Perhaps `session-startup-scripts@.service` or something like that. Having user-managerd involved there is just an implementation detail and not something that defines the service IMO.

src/sailfishusermanager.cpp
```diff
@@ -57,6 +57,8 @@ const int MAX_USERNAME_LENGTH = 20;
 const auto USER_ENVIRONMENT_DIR = QStringLiteral("/home/.system/var/lib/environment/%1");
 const auto USER_REMOVE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/remove.d");
 const auto USER_CREATE_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/create.d");
+const auto USER_PRE_SWITCH_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/pre-switch.d");
+const auto USER_SESSION_SCRIPT_DIR = QStringLiteral("/usr/share/user-managerd/session.d");
```
Tomin1: And of course if directories are adjusted then these need also adjusting.

Tomin1 (changes requested):
I'm still not quite convinced that this is the place to add more scripts for user sessions. The most it bothers me that with this change we are starting user-managerd also on device bootup adding even more work for the device when it starts up. Perhaps having these in the same repository but making the implementation less heavy could be an option. Anyway I need to discuss this with others.

The very least I would require for this to be merged is that condition on having some scripts to start user-managerd and then get that naming of the directories right. Apart from that directory name most things I commented are just implementation details that can be improved afterwards as well (but I still would not approve or merge unfinished work myself without any guarantee that it will get finished).

rinatdobr:
I've removed everything that's related to the `session`, and kept pre-switch state. I believe there is no need to rename that state.

Tomin1 (approved):
This should be okay to do. LGTM

PR #44, opened 2022-08-09
[user-managerd] Fix user id range and avoid running removeUserFiles for system users. JB#58459

krnlyng:


src/sailfishusermanager.cpp
```diff
@@ -461,8 +461,12 @@ void SailfishUserManager::setUserLimits(uint uid)
 int SailfishUserManager::removeUserFiles(const char *user)
 {
     struct passwd *pwd = getpwnam(user);
-    if (pwd)
-        return removeUserFiles(pwd->pw_uid);
+    if (pwd) {
+        if (pwd->pw_uid >= OWNER_USER_UID && pwd->pw_uid <= OWNER_USER_UID + MAX_RESERVED_UID) {
```
Tomin1: I would maybe define a new variable for the maximum UID instead of `OWNER_USER_ID + MAX_RESERVED_UID`. Maybe `MAX_USER_UID` or something, feel free to suggest.

Additionally maybe this could be on a single line with the previous `if`:
```
if (pwd && pwd->pw_uid >= OWNER_USER_UID && pwd->pw_uid <= MAX_USER_UID) {
```

Tomin1: And then could also add a static assert to the beginning that `SAILFISH_USERMANAGER_GUEST_UID` is within this limit.

Tomin1: I guess technically the strictest we could limit this to is `[OWNER_USER_UID, OWNER_USER_UID + SAILFISH_USERMANAGER_MAX_USERS[ | SAILFISH_USERMANAGER_GUEST_UID`, but it's debatable what's better. :smiley:

Tomin1 (approved):
LGTM now, let's check if Simo has some comment or not.

And just to clarify: It's also fine by me to merge this without his approval. He might not have time to check this.

spiiroin (approved):

mkosola (approved):

